PR #10, opened 2020-07-31
SolutionDomTask1

Victor-Zinchenko-mate:


src/scripts/main.js
```diff
@@ -1,3 +1,16 @@
+/* eslint-disable no-unused-vars */
```
vpolets, line 1: видали цей рядок

Victor-Zinchenko-mate: 'use strict';

const list = document.querySelectorAll('.population');

const text = [...list].map(item => +item.innerHTML.split(',').join(''));
const totalSum = text.reduce((a, b) => a + b, 0);
const averageSum = text.reduce((a, b) => a + b / text.length, 0).toFixed(0);

totalPopulation.textContent = totalSum;

averagePopulation.textContent = averageSum;

PR #28, opened 2020-08-21
solution

furude-rika:
- [DEMO LINK](https://furude-rika.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -3,7 +3,7 @@
 const populations = document.querySelectorAll('.population');
 
 const arrayOfPopulations = [...populations]
-  .map(text => text.innerText)
+  .map(elem => elem.innerText)
   .map(elem => elem.split(',').join(''));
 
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,17 @@
 'use strict';
 
-// write your code here
+const span = document.querySelectorAll('.population');
```
irenhh, line 3: it's not span

src/scripts/main.js
```diff
@@ -1,3 +1,17 @@
...
+
+const arr = [...span]
```
irenhh, line 5: what arr?

src/scripts/main.js
```diff
@@ -1,3 +1,17 @@
...
+  .map(text => text.innerText)
```
irenhh, line 6: it's not text

irenhh (changes requested):
image
would be nice to convert them according to the format in the table

irenhh (approved):

PR #43, opened 2020-08-24
completed

DmJSCourse:
https://dmjscourse.github.io/js_get_data_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,14 @@
 'use strict';
 
-// write your code here
+const popsCollection = document.getElementsByClassName('population');
+const popsArray = [];
+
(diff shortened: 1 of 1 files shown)


JuliaNastenko (approved):

src/scripts/main.js
```diff
@@ -1,3 +1,14 @@
 'use strict';
 
-// write your code here
+const popsCollection = document.getElementsByClassName('population');
```
irenhh, line 3: what is pops?)

irenhh: don't shorten words please

irenhh (commented):
image
convert it, so that it looks like numbers in the table

irenhh (approved):

PR #133, opened 2020-12-29
first solution

Layrovell:
[DEMO LINK](https://Layrovell.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const container = document.querySelector('.container');
-
 const population = container.querySelectorAll('.population');
 
 const arr = [...population].map(item => item.innerText);
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,23 @@
 'use strict';
 
-// write your code here
+const container = document.querySelector('.container');
+
+const population = container.querySelectorAll('.population');
+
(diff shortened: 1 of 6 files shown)


Serhii-Alieksieievych (approved):

src/scripts/main.js
```diff
@@ -1,3 +1,22 @@
 'use strict';
 
-// write your code here
+const container = document.querySelector('.container');
```
varseniuk, line 3:
```suggestion
const population = document.querySelectorAll('.population');
const totalPopulation = document.querySelector('.total-population');
const averagePopulation = document.querySelector('.average-population');
```

You can easily get rid of this redundant 'container' and use 'document' instead. You're searching for specific class-names and there no ned to find inside 'container' in this task. Just FYI

src/scripts/main.js
```diff
@@ -1,3 +1,22 @@
...
+const population = container.querySelectorAll('.population');
+
+const arr = [...population].map(item => item.innerText);
+const arrNumbers = (arr.join(' ').replace(/,/g, '')
+  .split(' ')).map(a => +a);
```
varseniuk, line 8:
```suggestion
const arr = [...population].map(item => 
  +item.textContent
    .split(',')
    .join('')

varseniuk (approved):

PR #137, opened 2021-01-01
solution

khrulenko:


olesiatk (approved):

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
 'use strict';
 
-// write your code here
+const numberFormatting = function(n) {
+  const formatted = [];
+
+  if (String(n).length % 3 > 0) {
+    formatted.push(String(n).substr(0, String(n).length % 3));
+  }
```
vsm2k17, line 8: you can avoid this condition if you iterate through the array from the end

vsm2k17 (approved):

PR #141, opened 2021-01-03
solution

HaidaiAndrii:


sashalarin202 (approved):
great job!

src/scripts/main.js
```diff
@@ -1,3 +1,15 @@
 'use strict';
 
-// write your code here
+const totalPopulation = document.querySelector('.total-population');
+const items = document.querySelectorAll('.population');
```
vsm2k17, line 4: use something more specific than items and list

vsm2k17 (changes requested):
you need to add comas
image

PR #151, opened 2021-01-03
Population calculated

MaximZhuravlov:
- [DEMO LINK](https://MaximZhuravlov.github.io/js_get_data_DOM/)


xipholena (approved):
Well done!

varseniuk (changes requested):
You should return the number in the format 452,543,964 instead of 452 543 964.
Try to change locale to 'en-US'

varseniuk (approved):

PR #164, opened 2021-01-04
Develop

NazarDovhanchuk:
[DEMO LINK](https://nazardovhanchuk.github.io/js_get_data_DOM/)


zhenya-mezhueva (approved):

varseniuk (changes requested):
You should return the number in the format 452,543,964 instead of 452 543 964.
Try to change locale to 'en-US'

varseniuk (approved):

PR #168, opened 2021-01-04
solution

alexmogwaiii:


5Mountains (approved):
Good solution! 👍  Despite this, it might be better to use populationList.length value instead of a number - 10 when calculating the average value, because if you only look at the js code, it's not clear where the value of this number is taken from (:

alexmogwaiii:
> Good solution! 👍 Despite this, it might be better to use populationList.length value instead of a number - 10 when calculating the average value, because if you only look at the js code, it's not clear where the value of this number is taken from (:

Thank you!) I forgot about magic numbers :)

src/scripts/main.js
```diff
@@ -1,3 +1,14 @@
...
-// write your code here
+const populationList = document.querySelectorAll('.population');
+const totalPopulation = document.querySelector('.total-population');
+const averagePopulation = document.querySelector('.average-population');
+
+const total = [...populationList].reduce((sum, item) => {
+  return sum + Number(item.textContent.split(',').join(''));
+}, 0);
+
+const average = total / 10;
```
vsm2k17, line 11: use the length of an array instead of `10`

vsm2k17 (approved):

PR #259, opened 2021-05-14
added average an total

Ihor-Karpyn:
- [DEMO LINK](https://Ihor-Karpyn.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,37 @@
 'use strict';
 
-// write your code here
+let data = document.getElementsByClassName('population');
+
+data = [...data].map(tag =>
+  Number(tag.innerText.split(',').join('')),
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,37 @@
 'use strict';
 
-// write your code here
+let data = document.getElementsByClassName('population');
+
+data = [...data].map(tag =>
```
mykola-yarchuk, line 5: it's better if the type of your variable always the same, just create another variable for this

mykola-yarchuk (approved):
great

src/scripts/main.js
Ihor-Karpyn: Ok!

PR #299, opened 2021-08-14
main.js file modified. total and average ages added

ihor-jpeg:
- [DEMO LINK](https://ihor-jpeg.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,32 @@
 'use strict';
 
-// write your code here
+const threeDigitSeparator = (number) => {
+  const stringNumber = number.toString();
+  const result = [];
+
+  for (let i = 0; i < stringNumber.length; i++) {
+    if (i % 3 === 0) {
+      result.push(',');
+    }
+    result.push(stringNumber.split('').reverse()[i]);
+  }
+
+  return result.reverse().slice(0, -1).join('');
+};
+
+const populations = [
+  ...document.querySelectorAll('.population'),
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+    }
+    result.push(stringNumber.split('').reverse()[i]);
+  }
+
+  return result.reverse().slice(0, -1).join('');
+};
+
+const populations = [
+  ...document.querySelectorAll('.population'),
+].map(country => +country.textContent.replace(/,/g, ''));
```
Anna-Tkachenko, line 19: 👍

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
-// write your code here
+const threeDigitSeparator = (number) => {
+  const stringNumber = number.toString();
+  const result = [];
+
+  for (let i = 0; i < stringNumber.length; i++) {
+    if (i % 3 === 0) {
+      result.push(',');
+    }
+    result.push(stringNumber.split('').reverse()[i]);
```
Anna-Tkachenko, line 11: Maybe it's better to split and reverse stringNumber before cycle, instead of doing it in every iteration

Anna-Tkachenko (approved):
Nice Going👍

src/scripts/main.js
ihor-jpeg: yes, thank you! i don't know how i missed that one.

PR #307, opened 2021-08-16
The task has been done

Moroz-Dmytro:
- [DEMO LINK](https://Moroz-Dmytro.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,28 @@
 'use strict';
 
-// write your code here
+function thousandSeparator(number) {
+  let value = number;
+  let valueFormatted = '';
+
+  while (value > 0) {
+    valueFormatted = value % 1000 + ',' + valueFormatted;
+    value = Math.trunc(value / 1000);
+  }
+
+  return valueFormatted.slice(0, -1);
+}
+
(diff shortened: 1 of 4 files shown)


SashaLifashkin:
Хорошая работа!

src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
 'use strict';
 
-// write your code here
+function thousandSeparator(number) {
```
varseniuk, line 3: You could just use `toLocaleString` method instead of this function

src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
...
+  return valueFormatted.slice(0, -1);
+}
+
+const allPopulation = document.querySelectorAll('.population');
+
+const population = [...allPopulation].map(
+  item => +item.textContent.split(',').join('')
+);
+
+const totalValue = population.reduce((sum, x) => sum + x, 0);
```
varseniuk, line 21:
```suggestion
const totalValue = population.reduce((sum, x) => sum + x);
```

varseniuk (changes requested):
Good job, but let's get rid of `thousandSeparator` function

PR #322, opened 2021-09-24
solution

Bohdan-Kochetov:


src/scripts/main.js
```diff
@@ -1,3 +1,24 @@
 'use strict';
 
 // write your code here
+const population = document.querySelectorAll('.population');
+const totalPopul = document.querySelector('.total-population');
+const averagePopul = document.querySelector('.average-population');
+const value = [...population].map(item => +item.innerText.split(',').join(''));
+const total = value.reduce((a, b) => a + b);
+const average = total / 10;
```
Anna-Tkachenko, line 9: What is 10? Use a length of values here

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):

PR #471, opened 2022-06-20
Solution

stanislav-topikha:


src/scripts/main.js
```diff
@@ -1,3 +1,14 @@
 'use strict';
 
-// write your code here
+let populationsList = document.getElementsByClassName('population');
```
illiarerun, line 3: why `let`? don't change value of variables if it's unnecessary
```suggestion
const populationsList = document.getElementsByClassName('population');
```

src/scripts/main.js
```diff
@@ -1,3 +1,14 @@
...
+const total = document.querySelector('.total-population');
+const average = document.querySelector('.average-population');
+
+populationsList = [...populationsList]
+  .map(value => +value.textContent.split(',').join(''));
```
illiarerun, line 8: it may lead to unexpected issues, so just create another var here

illiarerun (changes requested):
GJ, little fix left:

src/scripts/main.js
stanislav-topikha: fixed

src/scripts/main.js
stanislav-topikha: fixed

illiarerun (approved):

PR #515, opened 2022-07-21
Added total and average population

LordAsmodey:
- [DEMO LINK](https://LordAsmodey.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,17 @@
 'use strict';
 
-// write your code here
+const population = document.getElementsByClassName('population');
+
+const totalPopulationElement = document
+  .getElementsByClassName('total-population')[0];
+const averagePopulationElement = document
+  .getElementsByClassName('average-population')[0];
+
+const totalPopulation = [...population].reduce((acc, country) => {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,17 @@
 'use strict';
 
-// write your code here
+const population = document.getElementsByClassName('population');
+
+const totalPopulationElement = document
+  .getElementsByClassName('total-population')[0];
+const averagePopulationElement = document
+  .getElementsByClassName('average-population')[0];
```
solaryasha, line 8: You should use only `querySelector` or `querySelectorAll`
They are the most flexible

solaryasha (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,15 @@
 'use strict';
 
-// write your code here
+const population = document.getElementsByClassName('population');
```
solaryasha, line 3:
```suggestion
const population = document.querySelectorAll('.population');
```

solaryasha (approved):

PR #519, opened 2022-07-23
total & average population were calculated

PavloMarkov:


fledith (approved):
👍

src/scripts/main.js
```diff
@@ -1,3 +1,22 @@
...
+const avaragePopulation = totalPopulation / population.length;
+
+const totalToIncert = totalPopulation.toLocaleString('en-US');
+const avarageToIncert = avaragePopulation.toLocaleString('en-US');
+
+const elementTotal = document.querySelectorAll('.total-population');
+const elementAvarage = document.querySelectorAll('.average-population');
+
+elementTotal[0].innerHTML = totalToIncert;
+elementAvarage[0].innerHTML = avarageToIncert;
```
ivan-kadykalo, line 22: If you will use `.querySelector` instead of `.querySelectorAll`, you will get the only first element. In this case, you don't use an index in elementTotal[0], because you already will have only 1 element.

ivan-kadykalo (changes requested):

src/scripts/main.js
PavloMarkov: fixed

Ihor-Karpyn (approved):

ivan-kadykalo (approved):

PR #538, opened 2022-08-15
solution

Kulchevych:
- [DEMO LINK](https://Kulchevych.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,29 @@
 'use strict';
 
-// write your code here
+function normalizeNumber(str, separator) {
+  let normalizeStr = '';
+  const reversedStr = str.split('').reverse().join('');
+
+  for (let i = 0; i < reversedStr.length; i++) {
+    if (i % 3 === 0 && i !== 0) {
+      normalizeStr += separator;
+    }
+    normalizeStr += reversedStr[i];
+  }
+
+  return normalizeStr.split('').reverse().join('');
+}
+
+const countriesPopulation = document.querySelectorAll('.population');
+const total = [...countriesPopulation]
+  .map(item => +item.textContent.split(',').join(''))
+  .reduce((a, b) => a + b)
+  .toString();
+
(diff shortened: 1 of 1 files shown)


AlexanderKolomiiets (approved):
Nice Solution

src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
...
+  }
+
+  return normalizeStr.split('').reverse().join('');
+}
+
+const countriesPopulation = document.querySelectorAll('.population');
+const total = [...countriesPopulation]
+  .map(item => +item.textContent.split(',').join(''))
+  .reduce((a, b) => a + b)
+  .toString();
```
AlinaMatuschak, line 21: Do you need here convert to string?

```suggestion
const total = [...countriesPopulation]
  .map(item => +item.textContent.split(',').join(''))
  .reduce((a, b) => a + b);
```

src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
 'use strict';
 
-// write your code here
+function normalizeNumber(str, separator) {
```
AlinaMatuschak, line 3: you can use [toLocaleString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString)

src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
...
+    normalizeStr += reversedStr[i];
+  }
+
+  return normalizeStr.split('').reverse().join('');
+}
+
+const countriesPopulation = document.querySelectorAll('.population');
+const total = [...countriesPopulation]
+  .map(item => +item.textContent.split(',').join(''))
+  .reduce((a, b) => a + b)
```
AlinaMatuschak, line 20: Where is the default value for reduce?

AlinaMatuschak (changes requested):
Almost perfect )

PR #542, opened 2022-08-16
Task_get-data-dom

MarkShapka:
- [DEMO LINK](https://MarkShapka.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,22 @@
 'use strict';
 
-// write your code here
+const bodyElement = document.body;
+const spanElement = bodyElement.querySelectorAll('.population');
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,22 @@
 'use strict';
 
-// write your code here
+const bodyElement = document.body;
+const spanElement = bodyElement.querySelectorAll('.population');
+let total = 0;
+const arr = [];
```
Bohdan-Maksymiuk, line 6: what does this variable store?
it should be more meaningful name for variable, not just `arr`

Bohdan-Maksymiuk (approved):

PR #556, opened 2022-08-24
Task is done

ElinaBahirova:
https://elinabahirova.github.io/js_get_data_DOM

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,36 @@
 'use strict';
 
 // write your code here
+
+let populations = document.querySelectorAll('span.population');
+
+populations
+  = [...populations].map(
+    population => +population.innerText.split(',')
+      .join(''));
+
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,36 @@
...
+    const part = Math.trunc(n / 1000);
+
+    result.push(n - part * 1000);
+    n = part;
+  } while (n > 1000);
+
+  result.push(n);
+
+  return result.reverse().join(',');
+}
```
stanislav-topikha, line 33: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString
you don't need this
just use `toLocaleString()`

stanislav-topikha (changes requested):
comments ⬇️

valerii-derkach (approved):

PR #577, opened 2022-09-11
solution

savchuksavchuk:
[DEMO LINK](https://savchuksavchuk.github.io/js_get_data_DOM/)


src/scripts/main.js
```diff
@@ -1,3 +1,55 @@
...
+    bilions += ',';
+  } else {
+    bilions = '';
+  }
+
+  if (milions >= 1) {
+    milions += ',';
+  } else {
+    milions = '';
+  }
```
SanyaBratashchuk, line 21: This is too complicated read about [Intl](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat)

src/scripts/main.js
```diff
@@ -1,3 +1,55 @@
...
+
+  if (hundreds < 100) {
+    hundreds = '0' + hundreds;
+  }
+
+  return `${bilions}${milions}${thousands}${hundreds}`;
+}
+
+const spans = document.querySelectorAll('.population');
```
SanyaBratashchuk, line 40:
```suggestion
const populations = document.querySelectorAll('.population');
```

src/scripts/main.js
```diff
@@ -1,3 +1,55 @@
...
+
+const numbers = [...spans].map(span => span.innerText);
+
+let total = 0;
+
+for (const number of numbers) {
+  total += +number.split(',').join('');
+};
```
SanyaBratashchuk, line 48: use reduce

SanyaBratashchuk (changes requested):

valerii-derkach (approved):

PR #583, opened 2022-09-17
Develop

MaxManis:


src/scripts/main.js
```diff
@@ -1,3 +1,20 @@
 'use strict';
 
 // write your code here
+const population = document.querySelectorAll('span.population');
+const totalElement = document.querySelector('span.total-population');
+const averageElement = document.querySelector('span.average-population');
+const populationNums = [];
+
+for (const populationElement of population) {
+  populationNums.push(+populationElement.innerText.split(',').join(''));
```
OleksiiBlezniuk, line 10: 1 - use `Number` instead of `+`, it's more readable
2 - use map instead of loop

src/scripts/main.js
```diff
@@ -1,3 +1,20 @@
...
+}
+
+const total = populationNums.reduce((a, b) => a + b);
```
OleksiiBlezniuk, line 13: use more meaningful variable names, in array methods too
It's hard to understand what is `a` and what is `b`

OleksiiBlezniuk (changes requested):
Solution for this task should be deployed

src/scripts/main.js
```diff
@@ -1,3 +1,19 @@
...
 // write your code here
+const population = document.querySelectorAll('span.population');
+const totalElement = document.querySelector('span.total-population');
+const averageElement = document.querySelector('span.average-population');
+const populationNums = [...population]
+  .map(populationElement => (
+    Number(populationElement.innerText.split(',').join(''))
+  ));
+
+const total = populationNums.reduce((prev, curr) => prev + curr);
```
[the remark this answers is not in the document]
OleksiiBlezniuk: That way is more correct 🙂  It's not a prev value as in `sort`, it's an accumulator

OleksiiBlezniuk (changes requested):
Don't increase selectors specificity if there is no need in it

OleksiiBlezniuk (approved):

PR #595, opened 2022-09-30
solution was added

Maksym845:
[DEMO LINK](https://Maksym845.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -13,11 +13,11 @@ function makeString(number) {
 }
 
 const population = document.querySelectorAll('.population');
-const total = [...population].map(item => {
+const total = population.map(item => {
   return Number(item.innerHTML.split(',').join(''));
 }).reduce((prev, item) => prev + item, 0);
 
-const average = total / [...population].length;
+const average = total / population.length;
 
 const totalSpan = document.querySelector('.total-population');
 const averageSpan = document.querySelector('.average-population');
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,26 @@
 'use strict';
 
-// write your code here
+function makeString(number) {
+  const numbers = String(number).split('');
+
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,26 @@
...
+    if (j % 3 === 0 && i !== 0) {
+      numbers[i] = ',' + numbers[i];
+    }
+  }
+
+  return numbers.join('');
+}
+
+const population = document.querySelectorAll('.population');
+const total = [...population].map(item => {
```
Denchepornyuk, line 16: you don't have to make a copy here, map is non-mutating method

src/scripts/main.js
```diff
@@ -1,3 +1,26 @@
...
+  return Number(item.innerHTML.split(',').join(''));
+}).reduce((prev, item) => prev + item, 0);
+
+const average = total / [...population].length;
```
Denchepornyuk, line 20: the same here, you don't need array copy here

Denchepornyuk (changes requested):

src/scripts/main.js
Maksym845: But here we get nodelist, with the spread operator I made the array from the nodelist.

src/scripts/main.js
```diff
@@ -1,3 +1,26 @@
...
+  const numbers = String(number).split('');
+
+  for (let i = numbers.length - 1, j = 1; i >= 0; i--, j++) {
+    if (j % 3 === 0 && i !== 0) {
+      numbers[i] = ',' + numbers[i];
+    }
+  }
+
+  return numbers.join('');
+}
```
OleksiiBlezniuk, line 13: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString

OleksiiBlezniuk (approved):

PR #603, opened 2022-10-08
Solution top Js_get_data_DOM

Fispil:


qoa11a (changes requested):
Provide demo link.

src/scripts/main.js
```diff
@@ -1,3 +1,16 @@
 'use strict';
 
 // write your code here
+const span = document.querySelectorAll('.population');
+
+const newFormat = new Intl.NumberFormat('en-Us');
+
+const ourList = [...span].map(item => +(item.innerText.split(',').join('')));
```
polosanya, line 8: What is ourList? I think it's better to provide more informative names

src/scripts/main.js
```diff
@@ -1,3 +1,16 @@
...
+
+const totalPopulation = ourList.reduce((item, sum) => sum + item, 0);
```
polosanya, line 10:
```suggestion
const totalPopulation = ourList.reduce((sum, item) => sum + item, 0);
```
On the contrary, first value is sum (accumulator). Also, I think item is not the best name for population number.

polosanya (changes requested):
Almost done! Please check my comments

src/scripts/main.js
```diff
@@ -5,12 +5,16 @@ const span = document.querySelectorAll('.population');
...
-const ourList = [...span].map(item => +(item.innerText.split(',').join('')));
+const countryList = [...span]
+  .map(item => +(item.innerText.split(',').join('')));
 
-const totalPopulation = ourList.reduce((item, sum) => sum + item, 0);
+const totalPopulation = countryList
+  .reduce((sum, population) => sum + population, 0);
 
 const totalSpan = document.querySelector('.total-population');
 const avarageSpan = document.querySelector('.average-population');
```
polosanya, line 15:
```suggestion
const averageSpan = document.querySelector('.average-population');
```

polosanya (approved):
Good job!

PR #607, opened 2022-10-09
Get data DOM

Dikey945:
[DEMO LINK](https://Dikey945.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,16 @@
 'use strict';
-
 // write your code here
+
+const populations = document.querySelectorAll('.population');
+const totalPopulation = document.querySelector('.total-population');
+const totalAverage = document.querySelector('.average-population');
+const populationSum = [...populations]
+  .map(item => item.innerText.split(',').join(''))
+  .reduce((sum, currentPopulation) => (
+    (+sum) + (+currentPopulation)
+  ), 0);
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,16 @@
 'use strict';
-
 // write your code here
+
+const populations = document.querySelectorAll('.population');
+const totalPopulation = document.querySelector('.total-population');
+const totalAverage = document.querySelector('.average-population');
+const populationSum = [...populations]
+  .map(item => item.innerText.split(',').join(''))
```
Andr1yK, line 8: do u really need to create a copy before `map`?

src/scripts/main.js
```diff
@@ -1,3 +1,16 @@
...
+  .reduce((sum, currentPopulation) => (
+    (+sum) + (+currentPopulation)
```
Andr1yK, line 10:
```suggestion
    Number(sum) + Number(currentPopulation)
```

Andr1yK (changes requested):

src/scripts/main.js
Dikey945: Yep, querrySelectorAll returns a pseudo array whose __proto__ property not containing array methods, that's why I before using map() convert it to array by destructuring

silchyn (approved):
Cool!

PR #608, opened 2022-10-10
cod for population country

RusNeday:


loralevitska (changes requested):
please deploy your demo link and add to description

AlisaVlasova (changes requested):
demo still doesn't work

src/scripts/main.js
```diff
@@ -1,3 +1,48 @@
...
+}
+
+const avaragePopulation = sumPopulationAllCountry
+/ arrayPopulationAllCountry.length;
+
+document.body
+  .children[0]
+  .children[2]
+  .children[0]
+  .innerHTML = divide(sumPopulationAllCountry);
```
TarasHoliuk, line 42: better get targets via `querySelector` - it's not clear what it means: 
```
body
  .children[0]
  .children[2]
  .children[0]
```

src/scripts/main.js
```diff
@@ -1,3 +1,48 @@
...
+}
+
+const populationAllCountry = document.getElementsByClassName('population');
+
+const arrayPopulationAllCountry = [...populationAllCountry].map(popular =>
+  popular.innerText);
+
+let sumPopulationAllCountry = 0;
+
+for (const countryPopulation of arrayPopulationAllCountry) {
```
TarasHoliuk, line 31: Using `for of` is ok, but would be nicer to use `reduce` here

TarasHoliuk (changes requested):

TarasHoliuk (approved):
:fire:

PR #610, opened 2022-10-12
Added task solution

oleksusov:


src/scripts/main.js
```diff
@@ -1,3 +1,13 @@
 'use strict';
 
-// write your code here
+const population = document.querySelectorAll('.population');
+const populationList = [...population].map(value =>
+  Number(value.innerHTML.replaceAll(',', '')));
+const total = populationList.reduce((prev, current) => prev + current, 0);
+const average = total / populationList.length;
```
polosanya, line 7:
```suggestion
const population = document.querySelectorAll('.population');
const populationList = [...population].map(value =>
  Number(value.innerHTML.replaceAll(',', '')));
  
const total = populationList.reduce((sum, current) => sum + current, 0);
const average = total / populationList.length;
```
It is better to name it `sum` because it's what method reduce returns, so it will be easier to understand how it works.

polosanya (changes requested):
Provide DEMO LINK for your task, so we can check how it looks
Also please check my comments)

oleksusov:
[DEMO LINK](https://github.com/oleksusov/js_get_data_DOM)

polosanya (approved):
Good job!

PR #612, opened 2022-10-15
Solution

JohnRust12:


alexandra-protyanova (changes requested):
add demo link to your pr

JohnRust12:
> 

- [DEMO LINK](https://JohnRust12.github.io/js_get_data_DOM/)

JohnRust12:
- [DEMO LINK](https://JohnRust12.github.io/js_get_data_DOM/)

Bohdan-Maksymiuk (changes requested):
same here, did you deploy changes?

src/scripts/main.js
```diff
@@ -1,3 +1,41 @@
...
+const totalPopArr = String(totalPop).split('').reverse();
+let totalPopComas = [];
+
+for (let n = 0; n < totalPopArr.length; n++) {
+  totalPopComas.push(totalPopArr[n]);
+
+  if ((n + 1) % 3 === 0 && n !== totalPopArr.length - 1) {
+    totalPopComas.push(',');
+  }
+};
```
Bohdan-Maksymiuk, line 18: looks like you used this logic twice - make common function for it

Bohdan-Maksymiuk: and try to rewrite logic with reduce

Bohdan-Maksymiuk (changes requested):

src/scripts/main.js
JohnRust12: It's impossible, averagePopArr.length !== totalPopArr.length. They couldn't be joined (extra ',' in the average.

loralevitska: why you need `averagePopArr.length` here? this comment was about your loop. You can process more easily totalPopArr with `reduce()` method.  

I guess this link can be useful for you
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce
https://stackoverflow.com/questions/61174562/how-reduce-in-javascript-work-if-initial-value-is-array

src/scripts/main.js
```diff
@@ -1,3 +1,41 @@
 'use strict';
 
-// write your code here
+const allPop = document.querySelectorAll('.population');
+const arrPop = [...allPop].map(number => +number.innerText.split(',').join(''));
```
loralevitska, line 4: `map()` returns new array, copy is redundant here

loralevitska: and try use more obvious convert in case with more operations than 1
bad `+numberString`
good `Number(numberString)`

loralevitska (changes requested):
please check all comments and avoid code duplicate

loralevitska (changes requested):
almost done! remove comments

JohnRust12:
> almost done! remove comments

Here it is....

src/scripts/main.js
```diff
@@ -1,3 +1,14 @@
 'use strict';
 
-// write your code here
+const allPop = document.querySelectorAll('.population');
+const arrPop = [...allPop].map(number =>
```
polosanya, line 4: `allPop` and `arrPop` look almost the same so can be easily confused in future.
Try to avoid such situations and provide concise, informative names for your variables.

polosanya (approved):
Good job! I'm approving it, but please pay attention to my comment

src/scripts/main.js
JohnRust12: Ok, got it

PR #615, opened 2022-10-18
Develop

NazikVr:
- [DEMO LINK](https://nazikvr.github.io/js_get_data_DOM/)


src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+  const country = chap.textContent.split('');
+
+  let temporaryStr = '';
+
+  for (const numb of country) {
+    if (numb !== ',') {
+      temporaryStr += numb;
+    }
+  }
+  total += +temporaryStr;
```
loralevitska, line 22: use obvious convert
```suggestion
  total += Number(temporaryStr);
```

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
 'use strict';
 
-// write your code here
+function numberWithCommas(x) {
+  return x.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ',');
+}
+
+const popu = document.querySelectorAll('.population');
```
loralevitska, line 7: try to use clear variable names

```suggestion
const population = document.querySelectorAll('.population');
```

loralevitska (changes requested):
not bad!
it can be better with improved code style
try to combine declarations of variables and don't leave empty lines between them, but separate logical blocks of code with an empty line

PR #618, opened 2022-10-20
solution

MelnychukYana:
https://MelnychulYana.github.io/js_get_data_DOM/


alexandra-protyanova (changes requested):
Your demo doesn't work, try to redeploy your project

PR #628, opened 2022-10-24
Total and average populations were calculated

mykyta-krainik:


src/scripts/main.js
```diff
@@ -1,3 +1,28 @@
...
+      .textContent
+      .replaceAll(',', '')
+    );
+
+    if (!isFinite(parsedPopulation)) {
+      return 0;
+    }
+
+    return parsedPopulation;
+  });
```
liliia-livandovskaya, line 15:
```suggestion
const populations = [...document.querySelectorAll('.population')];
const normalizePopulation = [...population].map((text) =>
	    parseInt(text.innerText.split(",").join(""))
	  );
 
```
can be simplified

liliia-livandovskaya (changes requested):
Great job! Let's rewrite it a little bit

liliia-livandovskaya (approved):

PR #631, opened 2022-10-29
total and average population by dom was created

iPanda177:
[DEMO LINK](https://iPanda177.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,16 @@
 'use strict';
 
-// write your code here
+const population = document.querySelectorAll('.population');
+
+const text = [...population].map(data =>
+  parseInt(data.textContent.split(',').join('')));
+
+const total = text.reduce((acc, cur) => acc + cur, 0);
+
+const average = total / text.length;
+
+const totalResult = document.querySelector('.total-population');
+const averageResult = document.querySelector('.average-population');
+
+totalResult.textContent = total.toLocaleString('en-Us');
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,16 @@
 'use strict';
 
-// write your code here
+const population = document.querySelectorAll('.population');
+
+const text = [...population].map(data =>
+  parseInt(data.textContent.split(',').join('')));
```
silchyn, line 6:
```suggestion
const text = population.map(data => (
  parseInt(data.textContent.split(',').join(''))
));
```

silchyn (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,17 @@
 'use strict';
 
-// write your code here
+const population = document.querySelectorAll('.population');
+
+const text = [...population].map(data => (
```
silchyn, line 5:
```suggestion
const text = population.map(data => (
```

silchyn (changes requested):

src/scripts/main.js
iPanda177: it won't work
image
image

Anna-Tkachenko: it doesn't work because querySelectorAll returns the list of elements, which you need to convert in array

Anna-Tkachenko (approved):
Good job!

PR #645, opened 2022-11-07
Solution

dmytro-legerda:


src/scripts/main.js
```diff
@@ -1,3 +1,23 @@
 'use strict';
 
-// write your code here
+const population = document.getElementsByClassName('population');
```
TarasHoliuk, line 3:
```suggestion
const population = [...document.getElementsByClassName('population')];
```
And instead of creating a new array `text`, iterating with `for` loop, use reduce directly on `population`

src/scripts/main.js
```diff
@@ -1,3 +1,23 @@
...
+const averPopulation = document.getElementsByClassName('average-population');
```
TarasHoliuk, line 4:
```suggestion
const averagePopulation = document.getElementsByClassName('average-population');
```

TarasHoliuk (changes requested):

tetiana-ivanchenko (approved):

PR #674, opened 2022-11-29
Develop

diolna:
- [DEMO LINK](https://diolna.github.io/js_get_data_DOM/)


src/scripts/main.js
```diff
@@ -1,3 +1,86 @@
 'use strict';
 
-// write your code here
+class DecimalFormat {
```
vlad-demchuk, line 3: oh.., let's not use classes for dom tasks

src/scripts/main.js
```diff
@@ -1,3 +1,86 @@
...
+}
+average = total / span.length;
+
+const averageClass = document.getElementsByClassName('average-population')[0];
+
+const totalClass = document.getElementsByClassName('total-population')[0];
+
+const numform = new DecimalFormat('0,000,000,000.0');
+
+averageClass.innerText = numform.format(average);
```
vlad-demchuk, line 85: i think you can use `.toLocaleString('en-US')` to avoid this difficult `format` method

vlad-demchuk (changes requested):

src/scripts/main.js
diolna: Thank you. fixed

diolna:
Thank you. fixed

src/scripts/main.js
diolna: Thank you. fixed

PR #683, opened 2022-12-03
Solution

VadymKim:


src/scripts/main.js
```diff
@@ -1,3 +1,62 @@
 'use strict';
 
-// write your code here
+const collection = document.querySelectorAll('.population');
```
Oleksandra-Pasenchenko, line 3: This solution looks too complicated for this task, try to simplify it

PR #687, opened 2022-12-06
js_get_data_DOM

SerhiiVovchenko:
- [DEMO LINK](https://SerhiiVovchenko.github.io/js_get_data_DOM/)


src/scripts/main.js
```diff
@@ -1,3 +1,15 @@
 'use strict';
 
-// write your code here
+const textFromPopulation = document.querySelectorAll('.population');
```
liliia-livandovskaya, line 3:
```suggestion
const population = document.querySelectorAll('.population');
```

src/scripts/main.js
```diff
@@ -1,3 +1,15 @@
...
+const numberOfPopulation = [...textFromPopulation].map(number =>
```
liliia-livandovskaya, line 4:
```suggestion
const populationNumber = [...textFromPopulation].map(number =>
```

liliia-livandovskaya (approved):

PR #697, opened 2022-12-13
Solution

webdevw:


src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+const numbers = [...textData].map((elem) => +elem.innerHTML.replace(/,/gi, ''));
+
+const total = numbers.reduce(
+  (accumulator, current) => accumulator + current,
+  0
+);
+
+const average = total / numbers.length;
+
+function numberToString(number = 0) {
```
polosanya, line 16: Check out this method
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString
You can format the number easier with it

polosanya (changes requested):
Add a demo link to the PR description and please check the comment)

loralevitska (changes requested):
Oops, you forgot to add the demo link, we can't test the work without it
If your have some problems with deploy fill free to ask in fe_chat

polosanya (changes requested):
Here should be a Demo link. Please add it
image

polosanya (approved):
Good job!

PR #700, opened 2022-12-15
Solution

Deniskakaka:


Virass (changes requested):
Put it in a new variable
`population.reduce((acc, el) => acc + el, 0)`

And add demo link

PR #701, opened 2022-12-15
Solution

Vasyliy22:


src/scripts/main.js
```diff
@@ -1,3 +1,35 @@
...
+  numbersArr.push(+arr[i]);
+}
+
+const total = numbersArr.reduce((x, y) => {
+  return x + y;
+}, 0);
+
+const average = numbersArr.reduce((x, y) => {
+  return Math.round((x + y) / numbersArr.length);
+}, 0);
```
loralevitska, line 27: redundant method (in fact, this is another additional loop, which means additional calculation). Try to optimize your solution. Create variable and increment it in 19 line in previous loop

loralevitska (changes requested):
almost done)

Vasyliy22:
Лучше не мог придумать

loralevitska (approved):

PR #709, opened 2022-12-20
 JS get data DOM

partnersinbahamas:


src/scripts/main.js
```diff
@@ -1,3 +1,22 @@
 'use strict';
 
 // write your code here
+
+const element = document.querySelectorAll('.population');
```
TarasHoliuk, line 5:
```suggestion
const population = [...document.querySelectorAll('.population')];
```

It will be an array, so it will allow you to get rid of `count`

TarasHoliuk (changes requested):
Deploy your app

partnersinbahamas:
Why I can't use count? And if I make [...document] nothing not works

src/scripts/main.js
```diff
@@ -1,3 +1,22 @@
...
+let total = 0;
+let count = 0;
+
+for (let pop of population) {
+  pop = +pop.textContent.split(',').join('');
+
+  total += pop;
+
+  count++;
+}
```
tetiana-ivanchenko, line 16: use array methods, for that you should make copy, as Taras mentioned

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #715, opened 2022-12-24
Solution

offsset:


offsset:
https://offsset.github.io/js_get_data_DOM/

src/scripts/main.js
```diff
@@ -1,3 +1,22 @@
 'use strict';
 
-// write your code here
+let summa = 0;
```
vlad-demchuk, line 3: let's use another variable name. eg populations, totalPopulations and so on

src/scripts/main.js
```diff
@@ -1,3 +1,22 @@
...
+
+const result = document.querySelectorAll('li');
```
vlad-demchuk, line 5: the same

src/scripts/main.js
```diff
@@ -1,3 +1,22 @@
...
+  const number = +sum.innerText.split(',').join('');
+
+  summa += number;
+}
+
+const total = document.querySelector('.total-population');
+
+total.innerText = new Intl.NumberFormat('en-us').format(summa);
+
+const average = document.querySelector('.average-population ');
```
vlad-demchuk, line 19:
```suggestion
const averagePopulation = document.querySelector('.average-population ');
```

vlad-demchuk (changes requested):
Pay attention to variable names

src/scripts/main.js
tetiana-ivanchenko: fix it, pls
https://github.com/mate-academy/js_get_data_DOM/pull/715/files#r1056909992

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #716, opened 2022-12-25
top countries

Mykola-Busarhin:
[DEMO LINK](https://Mykola-Busarhin.github.io/js_get_data_DOM/)

Hi, there!!! Spent a lot of time figuring out how to convert a string to a number. For some reason, '+' doesn`t work in this case( I guess that`s because comma separator in the figures) Found a solution, but not sure if this is the best one) Is there another way to do it? Thanks in advance!


src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
 'use strict';
 
-// write your code here
+const numbers = document.querySelectorAll('.population');
+
+const populations = [...numbers].map((el) => el.innerText);
+
+const separatedNumbers = populations.join('/');
```
DarkMistyRoom, line 7:
```suggestion
const numbers = document.querySelectorAll('.population');
const populations = [...numbers].map((el) => el.innerText);
const separatedNumbers = populations.join('/');
```

DarkMistyRoom: no need to make so many empty lines. but it isn't a mistake, just a suggestion

DarkMistyRoom (changes requested):
Ya, you can do it more simpler
for example:
image

src/scripts/main.js
Mykola-Busarhin: Thanks) More elegant and twice shorter) The logic is the same though.

DarkMistyRoom (approved):

PR #726, opened 2022-12-29
solution

Yaroslav-Radchuk:
[DEMO LINK](https://Yaroslav-Radchuk.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,16 @@
 'use strict';
 
-// write your code here
+const spans = document.querySelectorAll('.population');
+let totalNum = 0;
+let count = 0;
+
+for (const span of spans) {
+  totalNum += (+span.textContent.split(',').join(''));
+  count++;
+};
+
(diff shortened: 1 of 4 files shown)


loralevitska (changes requested):
image
Add commas

src/scripts/main.js
```diff
@@ -1,3 +1,16 @@
 'use strict';
 
-// write your code here
+const spans = document.querySelectorAll('.population');
```
polosanya, line 3: Better to specify what exactly this value is

polosanya (changes requested):
Please convert the number to the required format
image
image

src/scripts/main.js
Anastasiia-Svintsova, line 3:
```suggestion
const population = document.querySelectorAll('.population');
```

Anastasiia-Svintsova (approved):

PR #730, opened 2023-01-02
solution

AnnyKoval:


Viktor-Kost (approved):
Well done!

As a recommendation - name variables that link to node element with a suffix "ref" or "el"
It will be easier to read
Like total -> totalRef 
average -> averageRef

PR #735, opened 2023-01-05
Solution

yana-kotsulym:
https://yana-kotsulym.github.io/js_get_data_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,24 @@
 'use strict';
 
-// write your code here
+const population = document.querySelectorAll('.population');
+const total = document.querySelector('.total-population');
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,24 @@
...
+const total = document.querySelector('.total-population');
+const average = document.querySelector('.average-population');
+
+let numericPopulation = [...population].map(number => number.textContent);
+
+numericPopulation = numericPopulation.map(number => {
+  const arr = number.split(',');
+
+  return arr.join('');
+})
```
SanyaBratashchuk, line 13: no need to call map twice

SanyaBratashchuk (approved):

PR #739, opened 2023-01-06
Develop

skomskiy:
- [DEMO LINK](https://skomskiy.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -3,23 +3,23 @@
 const content = document.querySelectorAll('.population');
 let total = 0;
 let avarage = 0;
-const arr = [];
+const populationArray = [];
 
 for (const child of content) {
   const re = /,/gi;
 
-  arr.push(+child.textContent.replace(re, ''));
-
-  total = arr.reduce((a, b) => {
-    return a + b;
-  });
+  populationArray.push(+child.textContent.replace(re, ''));
 };
+
+total = populationArray.reduce((a, b) => {
+  return a + b;
+});
 
 avarage = total / content.length;
 
-const numFor = Intl.NumberFormat('en-US');
-const newForTotal = numFor.format(total);
-const newForAvarage = numFor.format(avarage);
+const numberFormat = Intl.NumberFormat('en-US');
+const formattedTotal = numberFormat.format(total);
+const formattedAvarage = numberFormat.format(avarage);
 
 const totalFinal = document.querySelector('.total');
 const averageFinal = document.querySelector('.average');
@@ -28,7 +28,7 @@ document.querySelector('.total-population').remove();
 
 totalFinal.insertAdjacentHTML('beforeend', `
   <span class="total-population">
-    ${newForTotal}
+    ${formattedTotal}
   </span>
 `);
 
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,38 @@
 'use strict';
 
-// write your code here
+const content = document.querySelectorAll('.population');
+let count = 0;
```
SanyaBratashchuk, line 4: u can use `lenght` property instead of this count

src/scripts/main.js
```diff
@@ -1,3 +1,38 @@
...
+let avarage = 0;
+
+for (const child of content) {
+  const re = /,/gi;
+
+  total += parseInt(child.textContent.replace(re, ''));
+  count++;
+};
+
+avarage = total / count;
```
SanyaBratashchuk, line 15: use reduce

SanyaBratashchuk (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,41 @@
...
+  total = arr.reduce((a, b) => {
+    return a + b;
+  });
+};
+
+avarage = total / content.length;
+
+const numFor = Intl.NumberFormat('en-US');
+const newForTotal = numFor.format(total);
+const newForAvarage = numFor.format(avarage);
```
ViktorHavryliuk, line 22:
```suggestion
const numberFormat = Intl.NumberFormat('en-US');
const formattedTotal = numberFormat.format(total);
const formattedAvarage = numberFormat.format(avarage);
```

src/scripts/main.js
```diff
@@ -1,3 +1,41 @@
...
+const arr = [];
+
+for (const child of content) {
+  const re = /,/gi;
+
+  arr.push(+child.textContent.replace(re, ''));
+
+  total = arr.reduce((a, b) => {
+    return a + b;
+  });
```
ViktorHavryliuk, line 15: You can move this outside of for loop. Use loop to fill the values array, and then when loop is done calculate total one time, instead of recalculating it each iteration

src/scripts/main.js
```diff
@@ -1,3 +1,41 @@
 'use strict';
 
-// write your code here
+const content = document.querySelectorAll('.population');
+let total = 0;
+let avarage = 0;
+const arr = [];
```
ViktorHavryliuk, line 6: It's good to make var names more representative, so that it's not needed to look up what you put in this var later to figure out what is it supposed to hold
```suggestion
const populationArray = [];
```

ViktorHavryliuk (changes requested):

skomskiy:
- [DEMO LINK](https://skomskiy.github.io/js_get_data_DOM/)

src/scripts/main.js
```diff
@@ -1,3 +1,41 @@
...
+const content = document.querySelectorAll('.population');
+let total = 0;
+let avarage = 0;
+const populationArray = [];
+
+for (const child of content) {
+  const re = /,/gi;
+
+  populationArray.push(+child.textContent.replace(re, ''));
+};
```
DarkMistyRoom, line 12: would be great to move converting logic to a separate function

DarkMistyRoom: or we can find a much shorter solution
image

DarkMistyRoom (approved):

PR #748, opened 2023-01-11
Total and average population values were added

anna-inozemets:
[DEMO LINK](https://anna-inozemets.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,32 @@
 'use strict';
 
-// write your code here
+function thousandsSeparator(number) {
+  const splitedNum = String(number).split('');
+  const modificatedNum = [];
+  let counter = 0;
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
...
+
+    if (counter % 3 === 0 && i !== 0) {
+      modificatedNum.push(',');
+    }
+  }
+
+  return modificatedNum.reverse().join('');
+}
+
+const populationsList = document.querySelectorAll('.population');
```
Anna-Tkachenko, line 20: if we know that we'll have only one list, will be better to use simple `querySelector`

Anna-Tkachenko (approved):
Good job! ⭐️

PR #756, opened 2023-01-15
all good

Mikhaylo-Orlov:
[DEMO LINK](https://Mikhaylo-Orlov.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,16 @@
 'use strict';
 
-// write your code here
+const spans = document.querySelectorAll('.population');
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,16 @@
 'use strict';
 
-// write your code here
+const spans = document.querySelectorAll('.population');
```
vlad-demchuk, line 3: use clearer variable names. eg `population` and so on

vlad-demchuk (approved):

PR #757, opened 2023-01-15
Develop

andrieieva-yuliia:
- [DEMO LINK](https://andrieieva-yuliia.github.io/js_get_data_DOM/)


NVyrsta (approved):
Nice job :)

src/scripts/main.js
```diff
@@ -1,3 +1,12 @@
 'use strict';
 
-// write your code here
+const listPopulation = document.querySelectorAll('.population');
```
Anna-Tkachenko, line 3: there will be only one main list with this class
we can use simple `querySelector` instead of getting the collection of some elements

src/scripts/main.js
```diff
@@ -1,3 +1,12 @@
...
+const listNumber = [...listPopulation].map(item =>
+  +item.textContent.split(',').join(''));
```
Anna-Tkachenko, line 5:
```suggestion
const listNumber = [...listPopulation].map(item => (
  +item.textContent.split(',').join('')
));
```

src/scripts/main.js
```diff
@@ -1,3 +1,12 @@
...
+const total = listNumber.reduce((a, b) => a + b, 0);
+
+document.querySelector('.total-population').replaceWith(
+  total.toLocaleString('en-US'));
```
Anna-Tkachenko, line 9: such code isn't clear for reading
will be better to create separated variable for refactored total value and then use this variable here in `replaceWith` method

src/scripts/main.js
```diff
@@ -1,3 +1,12 @@
...
+
+document.querySelector('.average-population').replaceWith(
+  (total / listPopulation.length).toLocaleString('en-US'));
```
Anna-Tkachenko, line 12: same here, such code isn't clear
```suggestion
document.querySelector('.average-population')
  .replaceWith(
    (total / listPopulation.length).toLocaleString('en-US')
  );
```

Anna-Tkachenko (approved):

PR #764, opened 2023-01-18
done

Aleksey-Aleksey:
[DEMO LINK](https://Aleksey-Aleksey.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,19 @@
 'use strict';
 
-// write your code here
+const population = document.querySelectorAll('.population');
+
+const allArrInOne = [...population].map(el => el.textContent);
+
+const convToNum = allArrInOne.map(el => +el.split(',').join(''));
+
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,19 @@
 'use strict';
 
-// write your code here
+const population = document.querySelectorAll('.population');
+
+const allArrInOne = [...population].map(el => el.textContent);
+
+const convToNum = allArrInOne.map(el => +el.split(',').join(''));
```
tetiana-ivanchenko, line 7:
```suggestion
const allArrInOne = [...population].map(el => +el.textContent.split(',').join(''));
```

tetiana-ivanchenko (changes requested):

src/scripts/main.js
Aleksey-Aleksey: fixed

tetiana-ivanchenko (approved):
Good job, just don't cut the variable names:
convertToNumber instead of convToNum

PR #765, opened 2023-01-20
DOM / js get data

valntyn:
**[DEMO LINK](https://valntyn.github.io/js_get_data_DOM/)**

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -69,7 +69,7 @@
         <li class="list__item">
           <div>
             <img src="images/flag-russia.gif" alt="Russia flag">
-            Russia:
+            undefined:
           </div>
           <span class="population">145,933,461</span>
         </li>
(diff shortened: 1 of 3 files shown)


src/index.html
```diff
@@ -69,7 +69,7 @@ <h1>TOP 10 LARGEST COUNTRIES BY POPULATION</h1>
         <li class="list__item">
           <div>
             <img src="images/flag-russia.gif" alt="Russia flag">
-            Russia:
+            undefined:
```
IvanFesenko, line 72: ))))

src/scripts/main.js
```diff
@@ -1,3 +1,18 @@
 'use strict';
 
-// write your code here
+const items = document.querySelectorAll('.population');
+
+const population = [...items].map(item =>
+  +(item.innerText.split(',').join('')));
```
IvanFesenko, line 6: No need to do spread, because map return new array

IvanFesenko (approved):
Good job. But check comment

PR #766, opened 2023-01-20
Finish task

U-Ra-All:


README.md
```diff
@@ -1,14 +1,15 @@
 1. Replace `<your_account>` with your Github username in the link
-    - [DEMO LINK](https://<your_account>.github.io/js_get_data_DOM/)
+   - [DEMO LINK](https://U-Ra-All.github.io/js_get_data_DOM/)
```
mykhalenych, line 2: also add in description

mykhalenych (changes requested):
link isn't working

src/scripts/main.js
```diff
@@ -1,3 +1,19 @@
-'use strict';
+"use strict";
 
 // write your code here
+const populationArray = [...document.querySelectorAll(".population")];
+
+const totalPopulation = populationArray.reduce(
+  (acc, p) =>
+    isNaN(parseInt(p.textContent.replace(/,/g, "")))
```
OlEzhA0, line 8: why do we need additional checks for `isNaN` and parseInt?
just use 
```
acc + +value.textContent.split(',').join('')
```

OlEzhA0 (changes requested):
Screenshot 2023-01-20 at 7 01 12 PM
your link is unavailable

src/scripts/main.js
U-Ra-All: Thanks, done!

Anastasiia-Svintsova (approved):

PR #776, opened 2023-02-01
solution

GolDenik:


src/scripts/main.js
```diff
@@ -1,3 +1,21 @@
 'use strict';
 
-// write your code here
+const populationSpans = document.body.querySelectorAll('.population');
+
+let populationSum = 0;
+
+populationSpans.forEach(element => {
+    populationSum += Number(element.textContent.replace(/,/g, ''));
+})
```
ArtemTopolskyi, line 9: lets practice 'reduce' method

src/scripts/main.js
```diff
@@ -1,3 +1,21 @@
...
+
+const totalElement = document.body.querySelector('.total');
+const averageElement = document.body.querySelector('.average');
```
ArtemTopolskyi, line 12: it's good practice to move all elements initializing to the start

ArtemTopolskyi (changes requested):

PR #780, opened 2023-02-04
Work done

petrolozynskyi:


src/scripts/main.js
```diff
@@ -1,3 +1,23 @@
...
+  const fromNumberToText = parseInt(node.innerText.replace(/\D/g, ''));
+
+  total += fromNumberToText;
+}
+
+const average = total / counter;
+
+const aver = document.body.querySelector('.average-population');
+
+aver.textContent = average.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ',');
```
SanyaBratashchuk, line 19: move number formatting to the separate function

src/scripts/main.js
```diff
@@ -1,3 +1,23 @@
 'use strict';
 
-// write your code here
+const element = document.body.getElementsByClassName('population');
+let total = 0;
+let counter = 0;
```
SanyaBratashchuk, line 5: why do you need this counter? You just can use length property

SanyaBratashchuk (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,24 @@
...
+
+for (const node of element) {
+  const fromNumberToText = parseInt(node.innerText.replace(/\D/g, ''));
+
+  total += fromNumberToText;
+}
+
+const average = total / element.length;
+
+const commas = function(number) {
```
SanyaBratashchuk, line 14:
```suggestion
const formatNumber = function(number) {
```

SanyaBratashchuk (approved):

PR #783, opened 2023-02-06
Solution

IrynaK-a:


src/scripts/main.js
```diff
@@ -1,3 +1,42 @@
 'use strict';
 
-// write your code here
+class ThousandsSeparator {
+  separate(number) {
```
agniya-i, line 4: Interesting solution! 
You can also use `toLocaleString();` method

agniya-i (approved):
Don't forget to add a demo link to your PR message

PR #784, opened 2023-02-07
total and average population calculated

kirill-vypirovskyi:
- [DEMO LINK](https://kirill-vypirovskyi.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,19 +1,16 @@
 'use strict';
 
-const array = [];
-const spans = document.querySelectorAll('.population');
+const populationSpans = document.querySelectorAll('.population');
 
-for (const population of spans) {
-  array.push(population.innerText);
-}
+const totalPopulation = [...populationSpans]
+  .reduce((sum, population) => {
+    return sum + Number(population.innerText.replace(/,/g, ''));
+  }, 0);
 
-const numbersArray = array.map(elem => +elem.replace(/,/g, ''));
-
-const total = numbersArray.reduce((a, b) => a + b, 0);
-const average = total / 10;
+const averagePopulation = totalPopulation / populationSpans.length;
 
 const totalSpan = document.querySelector('.total-population');
 const averageSpan = document.querySelector('.average-population');
 
-totalSpan.innerText = total.toLocaleString('en-US');
-averageSpan.innerText = average.toLocaleString('en-US');
+totalSpan.innerText = totalPopulation.toLocaleString('en-US');
+averageSpan.innerText = averagePopulation.toLocaleString('en-US');


src/scripts/main.js
```diff
@@ -1,3 +1,19 @@
 'use strict';
 
-// write your code here
+const array = [];
+const spans = document.querySelectorAll('.population');
```
OleksiiBlezniuk, line 4: I guess there can be a lot of spans on the page, let's specify what spans have you selected

src/scripts/main.js
```diff
@@ -1,3 +1,19 @@
...
+
+for (const population of spans) {
+  array.push(population.innerText);
+}
```
OleksiiBlezniuk, line 8: Let's use array iteration methods

OleksiiBlezniuk: Hint: NodeList that returned from querySelectorAll can be easily converted to array spread operator

src/scripts/main.js
```diff
@@ -1,3 +1,19 @@
...
+
+const numbersArray = array.map(elem => +elem.replace(/,/g, ''));
```
OleksiiBlezniuk, line 10: What numbers? Why array? Let's use more specific names for variables

src/scripts/main.js
```diff
@@ -1,3 +1,19 @@
...
+
+const total = numbersArray.reduce((a, b) => a + b, 0);
+const average = total / 10;
```
OleksiiBlezniuk, line 13: same here, total what? average what?

src/scripts/main.js
```diff
@@ -1,3 +1,19 @@
...
+
+const totalSpan = document.querySelector('.total-population');
+const averageSpan = document.querySelector('.average-population');
+
+totalSpan.innerText = total.toLocaleString('en-US');
+averageSpan.innerText = average.toLocaleString('en-US');
```
OleksiiBlezniuk, line 19: Nice one 👍

OleksiiBlezniuk (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,16 @@
...
+    return sum + Number(population.innerText.replace(/,/g, ''));
+  }, 0);
+
+const averagePopulation = totalPopulation / populationSpans.length;
+
+const totalSpan = document.querySelector('.total-population');
+const averageSpan = document.querySelector('.average-population');
+
+totalSpan.innerText = totalPopulation.toLocaleString('en-US');
+averageSpan.innerText = averagePopulation.toLocaleString('en-US');
```
sergii-nosachenko, line 16: IMHO, `textContent` is preferable. [That's why.](https://stackoverflow.com/a/35213639)

sergii-nosachenko (approved):
Good work 🚀
I think we need to add changing the flag to task requirements.

PR #791, opened 2023-02-11
add solution

NikitaNevv:
https://NikitaNevv.github.io/js_get_data_DOM/


src/scripts/main.js
```diff
@@ -1,3 +1,36 @@
...
+  let item = document.querySelectorAll('.population')[i].textContent;
+
+  populationArrey.push(item = +item.replaceAll(',', ''));
+}
+
+let totalPopulation = 0;
+
+populationArrey.map(sum => {
+  totalPopulation += sum;
+});
```
ViktorHavryliuk, line 16: It would be nice place to use reduce method
https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce

ViktorHavryliuk (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,36 @@
 'use strict';
 
-// write your code here
+const populationList = document.querySelectorAll('.population');
+const populationArrey = [];
+
+for (let i = 0; i < populationList.length; i++) {
+  let item = document.querySelectorAll('.population')[i].textContent;
+
+  populationArrey.push(item = +item.replaceAll(',', ''));
```
Volodymyr-Melnyk-Fron10der, line 9:
```suggestion
  populationArray.push(item = +item.replaceAll(',', ''));
```

src/scripts/main.js
```diff
@@ -1,3 +1,36 @@
...
+}
+
+let totalPopulation = 0;
+
+totalPopulation = populationArrey.reduce((accumulator, currentValue) => {
+  return accumulator + currentValue;
+});
```
Volodymyr-Melnyk-Fron10der, line 16: 1. I don't see you change totalPopulation somewhere else, so why not use const?
2. It`s better to set initial value for reduce
```suggestion
const totalPopulation = populationArrey.reduce((accumulator, currentValue) => {
  return accumulator + currentValue;
}, 0);
```

Volodymyr-Melnyk-Fron10der (changes requested):
Cool👍 But one moment to make this perfect

PR #796, opened 2023-02-14
the task was solved

YevheniiHura:
https://YevheniiHura.github.io/js_get_data_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,18 @@
 'use strict';
 
-// write your code here
+const element = document.body.children[0];
+const collectionOfNumbers = element.querySelectorAll('.population');
+const arrayOfNumbers = [...collectionOfNumbers].map(number => {
+  return +number.innerText.replaceAll(',', '');
+});
+
+const total = arrayOfNumbers.reduce((sum, current) => {
+  return sum + current;
+}, 0);
+const average = total / arrayOfNumbers.length;
+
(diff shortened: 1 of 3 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,18 @@
...
-// write your code here
+const element = document.body.children[0];
+const collectionOfNumbers = element.querySelectorAll('.population');
+const arrayOfNumbers = [...collectionOfNumbers].map(number => {
+  return +number.innerText.replaceAll(',', '');
+});
+
+const total = arrayOfNumbers.reduce((sum, current) => {
+  return sum + current;
+}, 0);
```
OlEzhA0, line 11:
```suggestion
const total = arrayOfNumbers.reduce((sum, current) =>
  sum + current
, 0);
```

src/scripts/main.js
```diff
@@ -1,3 +1,18 @@
 'use strict';
 
-// write your code here
+const element = document.body.children[0];
+const collectionOfNumbers = element.querySelectorAll('.population');
+const arrayOfNumbers = [...collectionOfNumbers].map(number => {
```
OlEzhA0, line 5: you don't copy your array when you use `.map`
```suggestion
const arrayOfNumbers = collectionOfNumbers.map(number => {
```

OlEzhA0 (approved):

PR #797, opened 2023-02-15
the task was solved

batiukova-nataliia:
- [DEMO LINK](https://batiukova-nataliia.github.io/js_get_data_DOM/)


src/scripts/main.js
```diff
@@ -1,3 +1,27 @@
 'use strict';
 
-// write your code here
+const population = document.getElementsByClassName('population');
+const arr = [];
+const texts = [...population].map(item => item.innerText);
```
OlEzhA0, line 5: why do we need copy array for map method
```suggestion
const texts = population.map(item => item.innerText);
```

src/scripts/main.js
```diff
@@ -1,3 +1,27 @@
...
+
+for (let i = 0; i < texts.length; i++) {
+  arr.push(+texts[i].split(',').join(''));
+}
+
+const sum = (prev, current) => {
+  return prev + current;
+};
+const total = arr.reduce(sum);
```
OlEzhA0, line 14:
```suggestion
const total = arr.reduce(sum, 0);
```

src/scripts/main.js
```diff
@@ -1,3 +1,27 @@
...
+const avg = Math.round(arr.reduce(sum) / arr.length);
+
+function convert(number) {
+  return Intl.NumberFormat('en-US').format(number);
+}
+
+const sumSpan = document.querySelector('.total-population');
+
+sumSpan.textContent = `${convert(total)}`;
```
OlEzhA0, line 23: it's should be already string
```suggestion
sumSpan.textContent = convert(total);
```

OlEzhA0 (changes requested):
hey! check please my small comments

src/scripts/main.js
batiukova-nataliia: We don't need copy of array, but without spread operator map doesn't work, as variable population isn't array, it's collection. So I transformed it into the array with spread operator, as it was shown in the video.

mykhalenych (approved):

PR #802, opened 2023-02-19
Solution

bojkovladislav:


src/scripts/main.js
```diff
@@ -1,3 +1,16 @@
...
+const totalValue = document.getElementsByClassName('total-population')[0];
+const averageValue = document.getElementsByClassName('average-population')[0];
+
+const list = [...textData]
+  .map(text => Number(text.textContent.split(',').join('')));
+
+const numberFormatter = Intl.NumberFormat('en-US');
+
+totalValue.textContent = list.reduce((a, b) => a + b, 0);
+averageValue.textContent = totalValue.textContent / list.length;
```
Volodymyr-Melnyk-Fron10der, line 13:
```suggestion
const total = list.reduce((a, b) => a + b, 0);
const average = totalValue.textContent / list.length;
```

src/scripts/main.js
```diff
@@ -1,3 +1,16 @@
...
+
+totalValue.textContent = numberFormatter.format(totalValue.textContent);
+averageValue.textContent = numberFormatter.format(averageValue.textContent);
```
Volodymyr-Melnyk-Fron10der, line 16:
```suggestion
totalValue.textContent = numberFormatter.format(total);
averageValue.textContent = numberFormatter.format(average);
```

Volodymyr-Melnyk-Fron10der (changes requested):
Hi. Great. But I don't get why you assigned value to DOM, it's better to make changes in DOM with correct, prepared data

PR #814, opened 2023-02-28
Develop

G37ua:


src/scripts/main.js
```diff
@@ -1,3 +1,18 @@
...
-// write your code here
+function countPopulation() {
+  const items = document.querySelectorAll('.population');
+
+  const nums = [...items].map(item =>
+    Number(item.innerText.split(',').join('')));
+  const totalPopulation = nums.reduce((a, b) => a + b, 0);
+  const averagePopulation = totalPopulation / nums.length;
+
+  document.querySelector('.total-population').innerText
```
agniya-i, line 11: not a big mistake, but it would be better to create variable for `document.querySelector('.total-population')`, and use variable here

agniya-i (approved):

PR #816, opened 2023-02-28
Get data DOM

ilona-romaniuk:
https://ilona-romaniuk.github.io/js_get_data_DOM/


agniya-i (changes requested):
Please redeploy demo, because it shows readme now

agniya-i (approved):

PR #827, opened 2023-03-02
Solution

Ihor-Zakharov:


src/scripts/main.js
```diff
@@ -1,3 +1,17 @@
 'use strict';
 
-// write your code here
+const list = document.querySelectorAll('.population');
+
+let sum = 0;
+let count = 0;
```
SanyaBratashchuk, line 6: why do you need count you just can use length

src/scripts/main.js
```diff
@@ -1,3 +1,17 @@
...
+
+for (const item of list) {
+  sum += Number(item.innerText.replaceAll(',', ''));
+  count++;
+}
```
SanyaBratashchuk, line 11: get sum via reduce method

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):

PR #832, opened 2023-03-06
solution

valeriy-mikheyev:
- [DEMO LINK](https:/valeriy-mikheyev.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,20 +1,16 @@
 'use strict';
 
-const allSpan = document.querySelectorAll('.population');
+const populationElements = document.querySelectorAll('.population');
 const sumSpan = document.querySelector('.total-population');
 const average = document.querySelector('.average-population');
 
 let sum = 0;
-let count = 0;
 
-for (const span of allSpan) {
+for (const span of populationElements) {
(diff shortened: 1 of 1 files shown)


valeriy-mikheyev:
[valeriy-mikheyev](https://valeriy-mikheyev.github.io/js_get_data_DOM/)

src/scripts/main.js
```diff
@@ -1,3 +1,20 @@
 'use strict';
 
-// write your code here
+const allSpan = document.querySelectorAll('.population');
```
slavanosatskyi, line 3: it is always a good idea to choose a name for a variable from domain area - in this case it can be populationElements or something like this, feel free to choose more suitable name

src/scripts/main.js
```diff
@@ -1,3 +1,20 @@
...
+const sumSpan = document.querySelector('.total-population');
+const average = document.querySelector('.average-population');
+
+let sum = 0;
+let count = 0;
```
slavanosatskyi, line 8: i think you can delete it. allSpan.length will do the trick

slavanosatskyi (changes requested):

vlad-demchuk (approved):

PR #833, opened 2023-03-08
Solution

Serhii-Shtokalo:


src/scripts/main.js
```diff
@@ -1,3 +1,23 @@
 'use strict';
 
-// write your code here
+const population = document.querySelectorAll('.population');
+const textData = [];
+
+population.forEach(item => {
+  textData.push(item.textContent);
+});
```
DarkMistyRoom, line 8:
```suggestion
const population = document.querySelectorAll('.population');
const textData = [...population];
```

DarkMistyRoom: it is the easy way to do the same

src/scripts/main.js
```diff
@@ -1,3 +1,23 @@
...
+
+let total = 0;
+
+textData.forEach(item => {
+  const num = item.split(',').join('');
+
+  total += Number(num);
+});
```
DarkMistyRoom, line 16: we can use reduce method which was created specially for this

DarkMistyRoom (changes requested):
Ooops! demo link doesn't work. try to redeploy the page. If you will have any concerns, feel free to ask in the chat
image

sTorba24 (changes requested):
Fix previous comments please

OleziO (approved):

PR #841, opened 2023-03-12
all fixed

8Melya8:
[DEMO LINK](https://8melya8.github.io/js_get_data_DOM/)


src/scripts/main.js
```diff
@@ -1,3 +1,13 @@
 'use strict';
 
-// write your code here
+const items = document.querySelectorAll('.population');
+
+const populations = [...items].map(item => item.innerText.split(',').join(''));
+
+const result = populations.reduce((sum, item) => sum + Number(item), 0);
```
sTorba24, line 7:
```suggestion
const result = populations.reduce((sum, item) => sum + +item, 0);
```

sTorba24 (changes requested):
Almost good! Seems your demo link is not working, try to run deploy again

src/scripts/main.js
8Melya8: I've tried deploying the demo links but it always throws an error. Tell me what to do?

PR #849, opened 2023-03-13
solution :)

katerynashylina:


src/scripts/main.js
```diff
@@ -1,3 +1,16 @@
...
-// write your code here
+const population = document.querySelectorAll('.population');
+const totalPopulation = document.querySelector('.total-population');
+const averagePopulation = document.querySelector('.average-population');
+
+let total = 0;
+
+for (const num of population) {
+  total += +num.innerText.replace(/,/g, '');
+}
```
DarkMistyRoom, line 11: ok, but also we can use reduce method

DarkMistyRoom (approved):

PR #860, opened 2023-03-17
Total and average population were calculated

serge-kontakt:
- [DEMO LINK](https://serge-kontakt.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,17 @@
 'use strict';
 
-// write your code here
+const pop = document.querySelectorAll('.population');
+
+const text = [...pop].map(item => item.innerText.replace(/,/g, ''));
+
+const number = text.map(item => Number(item));
+
+const sum = number.reduce((item, a) => item + a, 0);
+
+const average = sum / number.length;
+
+const popTot = document.querySelector('.total-population');
+const popAv = document.querySelector('.average-population');
+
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,17 @@
 'use strict';
 
-// write your code here
+const pop = document.querySelectorAll('.population');
+
+const text = [...pop].map(item => item.innerText.replace(/,/g, ''));
+
+const number = text.map(item => Number(item));
```
sTorba24, line 7:
```suggestion
const number = text.map(item => +item);
```

src/scripts/main.js
```diff
@@ -1,3 +1,17 @@
...
+
+const sum = number.reduce((item, a) => item + a, 0);
```
sTorba24, line 9:
```suggestion
const sum = number.reduce((sum, num) => sum + num, 0);
```

sTorba24: use more meaningful names, your 'item' is actually accumulator that will be a result of .reduce
second parameter is current element of array

sTorba24 (approved):
It's great! Little comments i left

src/scripts/main.js
serge-kontakt: Ok, thank you!

PR #874, opened 2023-03-25
Initial commit

RomanHrytsiuk:


src/scripts/main.js
```diff
@@ -1,3 +1,25 @@
...
+
+const arr = [...textCollection];
+const textArray = [];
+
+for (const item of arr) {
+  const textData = item.textContent;
+  const digitData = Number(textData.split(',').join(''));
+
+  textArray.push(digitData);
+}
```
SanyaBratashchuk, line 13: I case that you are not going to use array methods. There is no sense to make an Array from collection  
```suggestion
const textArray = [];

for (const item of textCollection) {
  const textData = item.textContent;
  const digitData = Number(textData.split(',').join(''));

  textArray.push(digitData);
}
```

src/scripts/main.js
```diff
@@ -1,3 +1,25 @@
 'use strict';
 
-// write your code here
+const textCollection = document.querySelectorAll('.population');
+
+const arr = [...textCollection];
+const textArray = [];
```
SanyaBratashchuk, line 6:
```suggestion
const normalizedData = [];
```

SanyaBratashchuk (changes requested):

polosanya (approved):
Good job!
Please don't forget to add a Demo link to the PR description, so we can properly check your work
image

PR #881, opened 2023-03-29
new version

Grechana:
https://grechana.github.io/js_get_data_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,24 @@
 'use strict';
 
+const populationsAll = document.body.children[0].children[1].children;
+let total = 0;
+let amountOfCountries = 0;
+
+for (const el of populationsAll) {
+  const population = (el.children[1].textContent);
+  const populationNum = +(population.replace(/,/g, ''));
+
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,24 @@
 'use strict';
 
+const populationsAll = document.body.children[0].children[1].children;
```
polosanya, line 3: You can use querySelectorAll here instead
https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelectorAll

src/scripts/main.js
```diff
@@ -1,3 +1,24 @@
...
+let total = 0;
+let amountOfCountries = 0;
+
+for (const el of populationsAll) {
+  const population = (el.children[1].textContent);
+  const populationNum = +(population.replace(/,/g, ''));
+
+  total += populationNum;
+  amountOfCountries += 1;
```
polosanya, line 12: You can also use populationsAll.length to get the amount of countries

polosanya (approved):
Good job!

PR #882, opened 2023-03-30
solution

daryna-hnidash:


src/scripts/main.js
```diff
@@ -1,3 +1,25 @@
...
+    const number = text.split(',').join('');
+
+    return +number;
+  });
+
+let total = 0;
+
+for (const elem of populationsArray) {
+  total += elem;
+}
```
IvanFesenko, line 17: use Array.reduce to calculate

IvanFesenko (changes requested):
Lets improve a bit

src/scripts/main.js
```diff
@@ -1,3 +1,20 @@
 'use strict';
 
-// write your code here
+const populations = document.querySelectorAll('.population');
+
+const populationsArray = [...populations]
+  .map(element => {
```
OlEzhA0, line 6: we don't need to create array copy for `map`

OlEzhA0 (changes requested):
hey! please deploy your project

loralevitska (approved):
👍

PR #887, opened 2023-04-01
Solution v.1.0

annperetiatko:
[DEMO LINK]
https://annperetiatko.github.io/js_get_data_DOM/


src/scripts/main.js
```diff
@@ -1,3 +1,22 @@
...
+for (const number of spanCollection) {
+  const textData = number.textContent;
+
+  allData.push(textData);
+  finalNumbersData = allData.map(num => Number(num.replace(/,/g, '')));
+}
+
+const total = finalNumbersData.reduce((sum, x) => sum + x, 0);
+const average = total / finalNumbersData.length;
+
```
etojeDenys, line 17: we can rewrite it much more simple
```
const total = [...spanCollection].reduce((prev, curr) => {
  return prev + +curr.textContent.split(',').join('');
}, 0);

const average = total / spanCollection.length;
```

etojeDenys (changes requested):

VitaliyBondarenko1982 (approved):

PR #888, opened 2023-04-01
Solution

Vaskovics:
https://vaskovics.github.io/js_get_data_DOM/


src/scripts/main.js
```diff
@@ -1,3 +1,18 @@
 'use strict';
 
-// write your code here
+const allSelectors = document.querySelectorAll('.population');
+
+let allData = [ ...allSelectors ].map(el => el.innerHTML);
+
+allData = allData.map(el => +el.split(',').join(''));
```
Volodymyr-Melnyk-Fron10der, line 7: you can merge those maps

Volodymyr-Melnyk-Fron10der (changes requested):
just one optimization recommendation

DarkMistyRoom (changes requested):
Good job, but you need to show data in the same format as was given
image

SerhiiKirik (approved):
Great 🔥

PR #892, opened 2023-04-03
solution

Oleh-Rybachuk:
https://oleh-rybachuk.github.io/js_get_data_DOM/


src/scripts/main.js
```diff
@@ -1,3 +1,27 @@
 'use strict';
 
-// write your code here
+const population = document.getElementsByClassName('population');
+const arr = [];
+const texts = [...population].map(item => item.innerText);
+
+for (let i = 0; i < texts.length; i++) {
+  arr.push(+texts[i].split(',').join(''));
+}
```
VitaliyBondarenko1982, line 9:
```suggestion
const arr = [...population].map(item => +item.innerText.split(',').join(''));
```
you can make all it in map iteration.
something like that

src/scripts/main.js
```diff
@@ -1,3 +1,27 @@
...
+
+const sum = (prev, current) => {
+  return prev + current;
+};
+const total = arr.reduce(sum);
+const avg = Math.round(arr.reduce(sum, 0) / arr.length);
```
VitaliyBondarenko1982, line 15:
```suggestion
const avg = Math.round(total / arr.length);
```
yoy already have total;

src/scripts/main.js
```diff
@@ -1,3 +1,27 @@
...
+const texts = [...population].map(item => item.innerText);
+
+for (let i = 0; i < texts.length; i++) {
+  arr.push(+texts[i].split(',').join(''));
+}
+
+const sum = (prev, current) => {
+  return prev + current;
+};
+const total = arr.reduce(sum);
```
VitaliyBondarenko1982, line 14:
```suggestion
const total = arr.reduce(sum, 0);
```

VitaliyBondarenko1982 (changes requested):

Viktor-Kost (approved):

PR #893, opened 2023-04-03
Solution

oksanatytanych:


src/scripts/main.js
```diff
@@ -1,3 +1,14 @@
...
-// write your code here
+const populationArray = document.querySelectorAll('[class=population]');
+const populationInCountries = [...populationArray].map(el =>
+  el.innerText.split(',').join(''));
+
+const sum = populationInCountries.reduce((res, el) => +el + res, 0);
+const average = sum / populationInCountries.length;
+
+document.querySelector('[class=total-population]').textContent
+  = new Intl.NumberFormat('ja-JP').format(sum);
```
etojeDenys, line 11: we can just use `toLocaleString()`

etojeDenys (changes requested):

etojeDenys (approved):
awesome

PR #897, opened 2023-04-06
solution

popeye88:


qoa11a (changes requested):
Please, provide working demo link.

popeye88:
It's so strange...
Checked twice, it my nick for sure. Maybe problem with node -v

PR #898, opened 2023-04-06
Done

Ilia991:


src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
 'use strict';
 
+const [...spanPop] = document.querySelectorAll('.population');
+
+const totalText = document.querySelector('.total-population');
+const averageText = document.querySelector('.average-population');
+
+const arrNumbers = spanPop.map(people =>
+  Number(people.innerText.split(',').join(''))
+);//
```
tetiana-ivanchenko, line 10:
```suggestion
);
```

src/scripts/main.js
```diff
@@ -1,3 +1,30 @@
...
+  const qwe = String(num).split('');
+
+  for (let i = qwe.length - 1; i >= 0; i -= 3) {
+    qwe[i - 3] += ',';
+  }
+
+  return qwe.join('');
+}
+
+totalText.innerText = addSeparator(totalPeople);
```
tetiana-ivanchenko, line 26: you can use .toLocaleString istead

tetiana-ivanchenko (changes requested):

src/scripts/main.js
Ilia991: i know i can use it toLocaleString but it puts spaces instead of commas

vlad-demchuk (approved):

PR #905, opened 2023-04-11
Solution

mxwolfy:


src/scripts/main.js
```diff
@@ -1,3 +1,15 @@
 'use strict';
 
-// write your code here
+const collection = document.querySelectorAll('[class="population"]');
```
mykhalenych, line 3: Instead of querying for all elements with class 'population' .population'

mykhalenych (changes requested):

src/scripts/main.js
mxwolfy: ok, thanks

src/scripts/main.js
```diff
@@ -1,3 +1,15 @@
...
+  .map(item => +(item.innerText.replaceAll(',', '')));
+
+const total = values.reduce((sum, x) => sum + x);
+const average = total / values.length;
+
+document.querySelector('.total-population').innerText
+  = total.toLocaleString();
+
+document.querySelector('.average-population').innerText
+  = average.toLocaleString();
```
Viktor-Kost, line 15: it looks much better if you move document.querySelector('.total-population') and document.querySelector('.average-population') to own variables

Viktor-Kost (approved):
Nice!
Left one recommendation

src/scripts/main.js
mxwolfy: understood, thanks

PR #909, opened 2023-04-13
Fedenko js_get_data_DOM

VladFedenko:
https://VladFedenko.github.io/js_get_data_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -2,13 +2,11 @@
 
 const population = document.querySelectorAll('.population');
 
-let total = 0;
-
-const callback = (x) => {
-  total += +(x.innerHTML.split(',').join(''));
+const callback = (prev, x) => {
+  return prev + +(x.innerHTML.split(',').join(''));
 };
 
-population.forEach(callback);
+const total = [ ...population ].reduce(callback, 0);
 
 const average = total / population.length;
 


src/scripts/main.js
```diff
@@ -1,3 +1,37 @@
 'use strict';
 
-// write your code here
+const population = document.querySelectorAll('.population');
+
+let total = 0;
+
+for (const countriePopulation of population) {
```
VitaliyBondarenko1982, line 7: Typo: change `countriePopulation` to `countryPopulation` for better readability.

VitaliyBondarenko1982: Consider using functional programming methods like reduce instead of a traditional loop.

src/scripts/main.js
```diff
@@ -1,3 +1,37 @@
...
+  const stringArray = countriePopulation.innerHTML.split(',');
+  let result = '';
+
+  for (const number of stringArray) {
+    result += number;
+  }
```
VitaliyBondarenko1982, line 13:
```suggestion
  const result = countriePopulation.innerHTML.split(',').join('');
```
looks like it can be simplier

src/scripts/main.js
```diff
@@ -1,3 +1,37 @@
...
+}
+
+const average = total / population.length;
+
+function toFormat(data) {
+  let resultString = data.toString();
+  const result = [];
+
+  do {
+    if (resultString.length % 3 !== 0) {
```
VitaliyBondarenko1982, line 25:
```suggestion
    if (resultString.length % 3) {
```
little shorter

VitaliyBondarenko1982 (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,32 @@
 'use strict';
 
-// write your code here
+const population = document.querySelectorAll('.population');
+
+let total = 0;
```
etojeDenys, line 5: Since you're using `forEach` to calculate the total population, you could use the `reduce` method instead to make the code more functional and concise. 😊

etojeDenys (changes requested):

OleziO (approved):

PR #910, opened 2023-04-14
dataDOM

theFallcon:
- [DEMO LINK](https://theFallcon.github.io/js_get_data_DOM/)


src/scripts/main.js
```diff
@@ -1,3 +1,20 @@
...
 
-// write your code here
+let total = 0;
+const data = document.querySelectorAll('.population');
+
+for (const country of data) {
+  const numbers = country.textContent.replace(/,/g, '');
+
+  total += Number(numbers);
+}
```
Anastasiia-Svintsova, line 10: You can replace it by the reduce method

Anastasiia-Svintsova (approved):
Try to use array methods instead of loops

PR #913, opened 2023-04-16
Solution

Zibi95:
[DEMO LINK](https://zibi95.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,35 @@
+/* eslint-disable max-len */
 'use strict';
 
-// write your code here
+function getNumbersFromNodes(nodeList) {
+  const numbersArray = [];
+
+  for (const node of nodeList) {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,35 @@
...
+  const numbersArray = [];
+
+  for (const node of nodeList) {
+    const number = +node.textContent.split(',').join('');
+
+    numbersArray.push(number);
+  }
+
+  return numbersArray;
+}
```
DorotaLeniecDev, line 14: You could easily count here total using this for-of loop instead of creating additional array. Simplify

DorotaLeniecDev (changes requested):
I like your code readability! Although you could make this code a bit simpler :)

Radoslaw-Czerniawski (approved):
Well done :)

PR #914, opened 2023-04-16
initial

MaksymKos:
https://maksymkos.github.io/js_get_data_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,22 +1,11 @@
 'use strict';
 
-const spans = document.querySelectorAll('.population');
+const populationList = document.querySelectorAll('.population');
 
-const populations = [];
+const totalPopulation = [...populationList].reduce((acc, item) =>
+  acc + parseInt(item.textContent.replace(/,/g, '')), 0);
 
-spans.forEach(span => {
-  populations.push(span.textContent);
-});
-
-const populationValue = [];
-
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
...
+  populations.push(span.textContent);
+});
+
+const populationValue = [];
+
+populations.forEach(population => {
+  populationValue.push(parseInt(population.replace(/,/g, '')));
+});
+
+const totalPopulation = populationValue.reduce((a, b) => a + b, 0);
```
VitaliyBondarenko1982, line 17:
```suggestion
const totalPopulation = [...spans].reduce((acc, item) => acc + parseInt(item.textContent.replace(/,/g, '')), 0);
```

you can make all it in reduce.
two forEach iterations redundant

src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
 'use strict';
 
-// write your code here
+const spans = document.querySelectorAll('.population');
```
VitaliyBondarenko1982, line 3:
```suggestion
const populationList = document.querySelectorAll('.population');
```
consider more meaningful naming

VitaliyBondarenko1982 (changes requested):

VitaliyBondarenko1982 (approved):

PR #917, opened 2023-04-18
done

victor-buldenko:


src/scripts/main.js
```diff
@@ -1,3 +1,29 @@
...
+function thousandsSeparator(num) {
+  const source = '' + num;
+  const result = [];
+
+  for (let i = source.length; i > 0; i -= 3) {
+    result.unshift(source.substring(i - 3, i));
+  }
+
+  return result.join(',');
+}
```
ivbutsykin, line 20: You can simple use `num.toLocaleString('en')` instead your method:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString

ivbutsykin (changes requested):

loralevitska (approved):
🔥

PR #928, opened 2023-04-24
Solution

oksana-skuba:
[Demo](https://oksana-skuba.github.io/js_get_data_DOM/)


AnnaShevchenkoQA (changes requested):
1. Use checklist for doing your homework
2. Do not change files that you don't need to change according your hw
3. Check the main.js file in your last commit
4. Check did you deployed a decision

PR #935, opened 2023-04-27
solution

mariuszgit:


src/scripts/main.js
```diff
@@ -1,3 +1,15 @@
 'use strict';
 
-// write your code here
+const totalOutput = document.querySelector('.total-population');
+const averageOutput = document.querySelector('.average-population');
+const populationCollection = document.querySelectorAll('.population');
+
+const populationArray = [...populationCollection].map(el => {
```
Radoslaw-Czerniawski, line 7: Instead of mapping these values into `populationArray` then reducing them, just reduce`[...populationCollection]` into `totalPopulation` here. Later to calculate averagePopulation  use `populationCollection`'s length

Radoslaw-Czerniawski (changes requested):
Well done 💪, one small refactor needed

Radoslaw-Czerniawski (approved):
Very good :D

PR #936, opened 2023-04-28
Solution

RomanOstrous:
- [DEMO LINK](https://RomanOstrous.github.io/js_get_data_DOM/)


src/scripts/main.js
```diff
@@ -1,3 +1,20 @@
...
+'use strict';
+
+const populationSpans = document.querySelectorAll('.population');
+let total = 0;
+
+populationSpans.forEach(span => {
+  const population = Number(span.textContent.split(',').join(''));
+  if (population) {
+    total += population;
+  }
```
Anastasiia-Svintsova, line 10: You could use the reduce method instead

Anastasiia-Svintsova (approved):

PR #937, opened 2023-04-30
solution

AnastasiiaVorobets:


AnastasiiaVorobets:
[DEMO LINK](https://AnastasiiaVorobets.github.io/js_get_data_DOM/)

src/scripts/main.js
```diff
@@ -1,3 +1,16 @@
 'use strict';
 
-// write your code here
+const population = document.querySelectorAll('.population');
+const total = document.querySelector('.total-population');
+const average = document.querySelector('.average-population');
+
+let totalPopulation = 0;
+
+[...population].map(countryPopulation => {
```
VitaliyBondarenko1982, line 9: use reduce method instead map. 
map return new changed array with items what return callback function, but we don't need new array and in your implementation map function return undefined.
```suggestion
conat totalPopulation = [...population].map((total, countryPopulation) => {
```

VitaliyBondarenko1982 (changes requested):

VitaliyBondarenko1982 (approved):

PR #952, opened 2023-05-09
add task solution

Vasyl-Zhyliakov:
https://Vasyl-Zhyliakov.github.io/js_get_data_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,38 @@
 'use strict';
 
-// write your code here
+const population = document.querySelectorAll('.population');
+const popValue = [];
+
+for (const character of population) {
+  popValue.push(getNumber(character.innerText));
+}
+
+function getNumber(string) {
+  let number = '';
+
+  for (let i = 0; i < string.length; i++) {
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,38 @@
...
+  for (let i = transform.length - 1; i >= 0; i--) {
+    if ((transform.length - i) % 3 === 0 && i !== 0) {
+      transform[i] = ',' + transform[i];
+    }
+  }
+
+  return transform.join('');
+}
+
+document.querySelector('.total-population').innerText = getNewFormat(total);
```
anstsot, line 37: you can use [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) in the future

anstsot (approved):
Great job, approved!

src/scripts/main.js
Vasyl-Zhyliakov: I already know about it, but it hasn't been in the lectures yet.

PR #963, opened 2023-05-12
solution

generatedman:
- [DEMO LINK](https://generatedman.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,27 @@
 'use strict';
 
-// write your code here
+let data = document.querySelectorAll('.population');
+const totalPopulation = document.querySelector('.total-population');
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,27 @@
...
+
+data = [...data].map((item) => {
+  const population = item.innerText;
+  let populationNumber = '';
+
+  for (const ch of population) {
+    if (ch !== ',') {
+      populationNumber += ch;
+    }
+  }
```
dinaminahi, line 15: Using for loop inside of  `map` is bad for perfomence.
Instead try using regex for population

dinaminahi (changes requested):

PR #965, opened 2023-05-12
add task solution

antonina-klishch:
https://antonina-klishch.github.io/js_get_data_DOM/


SanyaBratashchuk (changes requested):
result numbers should be formatted too 
image

SanyaBratashchuk (approved):

PR #971, opened 2023-05-15
 js_get_data_DOM_solution

YelyzavetaMolochynska:
[DEMO LINK](https://YelyzavetaMolochynska.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -3,11 +3,11 @@
 const populationList = document.querySelectorAll('.population');
 const populationArr = [...populationList].map(el =>
   +(el.innerText.split(',').join('')));
-const total = populationArr.reduce((a, b) => (a + b));
+const total = populationArr.reduce((acc, number) => (acc + number));
 const average = total / populationArr.length;
 
 document.querySelector('.total-population').textContent
-  = total.toLocaleString();
+  = total.toLocaleString('en-US');
 
 document.querySelector('.average-population').textContent
-  = average.toLocaleString();
+  = average.toLocaleString('en-US');


src/scripts/main.js
```diff
@@ -1,3 +1,13 @@
 'use strict';
 
-// write your code here
+const populationList = document.querySelectorAll('.population');
+const populationArr = [...populationList].map(el =>
+  +(el.innerText.split(',').join('')));
+const total = populationArr.reduce((a, b) => (a + b));
```
IvanFesenko, line 6: it's better to set default value for reduce method and use some more understandable names
```suggestion
const total = populationArr.reduce((acc, number) => (acc + number), 0);
```

src/scripts/main.js
```diff
@@ -1,3 +1,13 @@
...
+const average = total / populationArr.length;
+
+document.querySelector('.total-population').textContent
+  = total.toLocaleString();
```
IvanFesenko, line 10: Do not forget to set format you need
2023-05-15_16h30_16


```suggestion
  = total.toLocaleString('en-US');
```

src/scripts/main.js
```diff
@@ -1,3 +1,13 @@
...
+
+document.querySelector('.average-population').textContent
+  = average.toLocaleString();
```
IvanFesenko, line 13:
```suggestion
  = average.toLocaleString('en-US');
```

IvanFesenko (changes requested):
Also do not forget to deloy your work

IvanFesenko (approved):

PR #978, opened 2023-05-17
add task solution

maksym-nemera:
- [DEMO LINK](https://NickMitchell61.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,14 @@
 'use strict';
 
-// write your code here
+const allPopulation = document.querySelectorAll('.population');
+const totalPopulation = document.querySelector('.total-population');
+const averagePopulation = document.querySelector('.average-population');
+
+const result = [...allPopulation].map(
+  number => +number.textContent.replaceAll(',', ''))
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,14 @@
 'use strict';
 
-// write your code here
+const allPopulation = document.querySelectorAll('.population');
+const totalPopulation = document.querySelector('.total-population');
+const averagePopulation = document.querySelector('.average-population');
+
+const result = [...allPopulation].map(
```
dinaminahi, line 7: Instead of `result`, it's better to use something more descriptive like `totalPopulationCount`

src/scripts/main.js
```diff
@@ -1,3 +1,14 @@
...
+  number => +number.textContent.replaceAll(',', ''))
+  .reduce((sum, number) => sum + number, 0);
```
dinaminahi, line 9: and also here, instead of `number` for example `population`

dinaminahi (approved):
Great job 💪 

Just consider of using more descriptive names. I know it is difficult sometimes, but it worth it :)

PR #980, opened 2023-05-18
js_get_data_DOM

AndriiYelieva:
add task solution, but I doubt whether it's the right way


IvanFesenko (changes requested):
Check my comments

mykhalenych (approved):

mykhalenych (approved):

PR #988, opened 2023-05-24
js_get_data_DOM solution

RomanTsvindin:


agniya-i (changes requested):
The demo link isn't working now, please redeploy 🙌

OleziO (approved):

PR #995, opened 2023-05-26
Solution

yurii-shkrobut-m:
- [DEMO LINK](https://yurii-shkrobut-m.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,21 @@
 'use strict';
 
-// write your code here
+const populations = document.querySelectorAll('.population');
+let total = 0;
+
+for (const population of populations) {
+  const number = +population.textContent.split(',').join('');
+
+  total += number;
+}
+
+const average = total / populations.length;
+
+const averageFormatted = average.toLocaleString('en-US');
+const totalFormatted = total.toLocaleString('en-US');
+
+const averagePopulation = document.querySelector('.average-population');
+const totalPopulation = document.querySelector('.total-population');
+
+averagePopulation.textContent = averageFormatted;
+totalPopulation.textContent = totalFormatted;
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,21 @@
 'use strict';
 
-// write your code here
+const populations = document.querySelectorAll('.population');
+let total = 0;
+
+for (const population of populations) {
```
SerhiiKirik, line 6: Consider using `forEach` or `reduce` instead of a traditional for loop for iterating over the populations array.

src/scripts/main.js
```diff
@@ -1,3 +1,21 @@
...
+  const number = +population.textContent.split(',').join('');
```
SerhiiKirik, line 7: use `Number` instead `+`
```suggestion
  const number = Number(population.textContent.split(',').join(''));
```

src/scripts/main.js
```diff
@@ -1,3 +1,21 @@
 'use strict';
 
-// write your code here
+const populations = document.querySelectorAll('.population');
+let total = 0;
```
SerhiiKirik, line 4: Would be better to create a function & call this function to calculate a value

SerhiiKirik (changes requested):
Well done 🔥 
Let's improve your solution 🧑‍💻

src/scripts/main.js
```diff
@@ -1,3 +1,26 @@
...
+
+const calculateTotal = (num) => {
+  total += num;
+};
+
+const populations = document.querySelectorAll('.population');
+
+populations.forEach(el => {
+  const number = Number(el.textContent.split(',').join(''));
```
VitaliyBodnar, line 12: Consider using a more descriptive variable name, like 'populationNumber', instead of 'number'.

src/scripts/main.js
```diff
@@ -1,3 +1,26 @@
...
-// write your code here
+let total = 0;
+
+const calculateTotal = (num) => {
+  total += num;
+};
+
+const populations = document.querySelectorAll('.population');
+
+populations.forEach(el => {
```
VitaliyBodnar, line 11: 'el' it's not a correct name, try to provide a more descriptive name, like populations.forEach(population => {

IvanFesenko (approved):

PR #996, opened 2023-05-27
Added the solution

zynkd:
[DEMO LINK](https://zynkd.github.io/js_dom-get-data/)


src/scripts/main.js
```diff
@@ -1,3 +1,21 @@
 'use strict';
 
-// write your code here
+const populationElements = document.querySelectorAll('.population');
+
+const populationList = [...populationElements].map((element) => {
```
witflash, line 5: The same here. You can use directly `populationElements.map` because the `map` method returns a new array (this method doesn't mutate an original array)

witflash (approved):
Good job! 👍🏻

polosanya (approved):

PR #998, opened 2023-05-30
add task solution

sasdaiv:


VitaliyBondarenko1982 (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,17 @@
...
 
-// write your code here
+const popTags = document.querySelectorAll('.population');
+const items = [...popTags].map(element => {
+  const text = element.innerText;
+
+  return parseFloat(text.replace(/,/g, ''));
+});
+
+const totalPop = items.reduce((sum, currentVal) => sum + currentVal, 0);
```
polosanya, line 10:
```suggestion
const totalPopulation = items.reduce((sum, currentVal) => sum + currentVal, 0);
```
It is better to use full names

polosanya (approved):
Great!

PR #1004, opened 2023-06-02
solution

artem-ivashchenko:
DEMO LINK - https://artem-ivashchenko.github.io/js_get_data_DOM/

Question: How can i parse from .toLocaleString('en-US') (thousand separated commas) into a regular number without .split(',').join('')?


witflash (approved):
Well done 👍

PR #1007, opened 2023-06-03
solution

vlad-liakh:
https://vlad-liakh.github.io/js_get_data_DOM/


etojeDenys (changes requested):
your link does not work

vlad-liakh:
fixed a bug

OleziO (approved):

PR #1009, opened 2023-06-04
getDataDom

Tetiana-Hishchak:
https://Tetiana-Hishchak.github.io/js_get_data_DOM/


etojeDenys (changes requested):
we don't need "!"

etojeDenys (approved):
LGTM

PR #1012, opened 2023-06-05
Solution

Nataliia95254:


src/scripts/main.js
```diff
@@ -1,3 +1,21 @@
...
+
+for (let i = 0; i < list.length; i++) {
+  population.push(list[i].innerHTML.replaceAll(',', ''));
+}
+
+const total = population.reduce(function(sum, current) {
+  return sum + +current;
+}, 0);
+
+const average = total / 10;
```
demetrxx, line 15: what if we had more countries, or less? Replace 10 with a dynamic value

src/scripts/main.js
```diff
@@ -1,3 +1,21 @@
...
+const list = document.querySelectorAll('.population');
+
+const population = [];
+
+for (let i = 0; i < list.length; i++) {
+  population.push(list[i].innerHTML.replaceAll(',', ''));
+}
+
+const total = population.reduce(function(sum, current) {
+  return sum + +current;
```
demetrxx, line 12: It's better using explicit `Number('123')` method, as such option with 2 + operators with different functionalities inhibits code readability

demetrxx (approved):
Well done 👍 Go ahead!
Take note of the comments though and don't forget to pin demo link to description

PR #1017, opened 2023-06-06
solution

Gosyav:
https://vo7kov.github.io/js_get_data_DOM/

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,13 @@
 'use strict';
 
-// write your code here
+let population = document.querySelectorAll('.population');
+const totalText = document.querySelector('.total-population');
+const averageText = document.querySelector('.average-population');
+
+[...population] = [...population]
+  .map(number => +number.innerText.split(',').join(''));
+
+const sum = population.reduce((a, b) => a + b);
+
+totalText.innerText = sum.toLocaleString('en-US');
+averageText.innerText = (sum / population.length).toLocaleString('en-US');
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,13 +1,13 @@
 'use strict';
 
-let population = document.querySelectorAll('.population');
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,13 @@
...
 
-// write your code here
+let population = document.querySelectorAll('.population');
+const totalText = document.querySelector('.total-population');
+const averageText = document.querySelector('.average-population');
+
+[...population] = [...population]
+  .map(number => +number.innerText.split(',').join(''));
+
+const sum = population.reduce((a, b) => a + b);
```
Viktor-Kost, line 10: You need to provide an initial value for the reduce method.

src/scripts/main.js
```diff
@@ -1,3 +1,13 @@
 'use strict';
 
-// write your code here
+let population = document.querySelectorAll('.population');
+const totalText = document.querySelector('.total-population');
+const averageText = document.querySelector('.average-population');
+
+[...population] = [...population]
```
Viktor-Kost, line 7: Incorrect assignment to a NodeList. You should assign the mapped values to a new variable.

Viktor-Kost (changes requested):

src/scripts/main.js
```diff
@@ -1,3 +1,13 @@
...
 
-// write your code here
+const population = document.querySelectorAll('.population');
+const totalText = document.querySelector('.total-population');
+const averageText = document.querySelector('.average-population');
+
+const numbersArray = [...population]
+  .map(number => +number.innerText.split(',').join(''));
+
+const sum = numbersArray.reduce((a, b) => a + b);
```
VitaliyBondarenko1982, line 10: not fixed commetn to provide initial value to reduce.
of course in this case it is not change result, bun it is good practise to provide in any case

VitaliyBondarenko1982 (approved):

PR #1035, opened 2023-07-18
add task solution

Luk2asz:
- [DEMO LINK](https://Luk2asz.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,14 +1,12 @@
 'use strict';
 
 const population = document.getElementsByClassName('population');
-const populationStrings = [];
 let total = 0;
 
 for (let i = 0; i < population.length; i++) {
   const string = population[i].innerText.replace(new RegExp(',', 'g'), '');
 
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,20 @@
...
-// write your code here
+const population = document.getElementsByClassName('population');
+const populationStrings = [];
+let total = 0;
+
+for (let i = 0; i < population.length; i++) {
+  const string = population[i].innerText.replace(new RegExp(',', 'g'), '');
+
+  populationStrings.push(Number(string));
+  total += populationStrings[i];
```
DorotaLeniecDev, line 11: I think you dont need ```populationStrings``` array ( Why you call it strings if you are pushing numbers?). You could simply:
```suggestion
  total += Number(string);
```

DorotaLeniecDev (changes requested):
Check my comment

Viktor-Kost (approved):

PR #1037, opened 2023-07-21
js_get_data_DOM

KarolinaWiktoria:
https://KarolinaWiktoria.github.io/js_get_data_DOM/


src/scripts/main.js
```diff
@@ -1,3 +1,21 @@
 'use strict';
 
-// write your code here
+const population = document.querySelectorAll('.population');
+
+const content = Array.from(population).map(element => element.textContent);
+
+const contentNew = [];
+
+for (let i = 0; i < content.length; i++) {
```
darokrk, line 9: Instead of using a for loop to create a new array, you can try to use `map` method.

`const contentNew = content.map(item => Number(item.replace(/,/g, '')))`

src/scripts/main.js
```diff
@@ -1,3 +1,21 @@
...
+  contentNew.push(+content[i].replace(/,/g, ''));
```
darokrk, line 10: Maybe instead of `contentNew` name can be more descriptive like `populationArray` ?

darokrk (changes requested):
Nicely done 👏🏽  fix my mentions and we ready to go 🔥

PR #1039, opened 2023-08-01
sol

GGsmou:
[DEMO LINK](https://GGsmou.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,22 @@
 'use strict';
 
-// write your code here
+const populationValues = [...document.querySelectorAll('.population')]
+  .map(element => {
+    const populationText = element.textContent;
+
+    try {
+      return parseInt(populationText.replace(/,/g, ''), 10);
+    } catch (error) {
+      return 0;
+    }
+  });
+
+const totalPopulation = populationValues
+  .reduce((total, population) => total + population, 0);
+const averagePopulation = totalPopulation / populationValues.length || 0;
+
+const averagePopulationElement = document.querySelector('.average-population');
+const totalPopulationElement = document.querySelector('.total-population');
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,22 @@
...
+    try {
+      return parseInt(populationText.replace(/,/g, ''), 10);
+    } catch (error) {
+      return 0;
+    }
+  });
+
+const totalPopulation = populationValues
+  .reduce((total, population) => total + population, 0);
+const averagePopulation = totalPopulation / populationValues.length || 0;
```
OlEzhA0, line 16: why do we need to have this `|| 0`? check this condition in console `1 / 0 || 1`. You will see output `Infinity`, not 1, because when we will divide something to `0` it's be `Infinity` and this `|| 0` redundant and never works

OlEzhA0 (changes requested):
gj

src/scripts/main.js
```diff
@@ -1,3 +1,20 @@
...
+const populationValues = [...document.querySelectorAll('.population')]
+  .map(element => {
+    const populationText = element.textContent || '';
+
+    return parseInt(populationText.replace(/,/g, ''), 10) || 0;
+  });
+
+const totalPopulation = populationValues
+  .reduce((total, population) => total + population, 0);
+const averagePopulation = populationValues.length === 0
```
DanilWeda, line 12:
```suggestion
const averagePopulation = !populationValues.length
```

DanilWeda (approved):

src/scripts/main.js
DanilWeda: Good work! And check the test case, please.

PR #1059, opened 2023-10-16
add task solution

kbekher:
[DEMO LINK](https://kbekher.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,21 +1,19 @@
 'use strict';
 
-// write your code here
-function formatNumber(number) {
-  return number.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ',');
-}
-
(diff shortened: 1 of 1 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,21 @@
...
+elements.forEach(element => {
+  numbers.push(parseInt(element.innerHTML.replace(/,/g, '')));
+});
+
+const total = numbers.reduce((sum, number) => sum + number, 0);
+
+const average = total / numbers.length;
+
+document.querySelector('.total-population').innerHTML = formatNumber(total);
+document.querySelector('.average-population').innerHTML = formatNumber(average);
```
etojeDenys, line 21: instead of the `formatNumber` function, we can use `.toLocaleString('en-US')`

etojeDenys (changes requested):

PR #1074, opened 2023-11-17
add task solution

ANTI-bogdan:
[DEMO LINK](https://ANTI-bogdan.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,16 @@
 'use strict';
 
-// write your code here
+const allPopulationText = document.querySelectorAll('.population');
+const allPopulationNumber = [...allPopulationText].map(item => (
+  +item.innerText.replaceAll(',', '')));
+
+const totalPopulation = allPopulationNumber.reduce(
+  (prev, element) => prev + element);
+const averagePopulation = totalPopulation / allPopulationNumber.length;
+
+const total = document.querySelector('.total-population');
+const average = document.querySelector('.average-population');
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,16 @@
 'use strict';
 
-// write your code here
+const allPopulationText = document.querySelectorAll('.population');
+const allPopulationNumber = [...allPopulationText].map(item => (
```
VitaliyBodnar, line 4: - don't need to create a copy of allPopulationText as method map did it already 
```suggestion
const allPopulationNumber = allPopulationText.map(item => (
```
- item it's not a good name, try to provide more descriptive name

src/scripts/main.js
```diff
@@ -1,3 +1,16 @@
...
+  +item.innerText.replaceAll(',', '')));
+
+const totalPopulation = allPopulationNumber.reduce(
+  (prev, element) => prev + element);
```
VitaliyBodnar, line 8:
```suggestion
const totalPopulation = allPopulationNumber.reduce((prev, population) => {
  return prev + population;
});
```

VitaliyBodnar (changes requested):

src/scripts/main.js
ANTI-bogdan: it`s dont work. If doing like you write, it will be a collection, not an array (it haven`t map method)
Снимок экрана 2023-11-17 160247

anastasiiavorobiova (approved):
Great work!

PR #1122, opened 2024-02-18
add task solution

monikaprochniewicz:
https://monikaprochniewicz.github.io/js_get_data_DOM/


yevhenii-pyl (approved):
Good job!

I love your approach to variable naming: `totalPop `, `averagePop`, => works just fine, however, try to avoid obscure naming like `numbers || i || sum || avg`.

PR #1146, opened 2024-04-15
task complete

Manankin:
[DEMO LINK](https://manankin.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,38 @@
 'use strict';
 
-// write your code here
+const totalPopulation = document.querySelector('span.total-population');
+const averagePopulation = document.querySelector('span.average-population');
+const populationByCountries = document.querySelectorAll('span.population');
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,38 @@
...
+
+function getTotalPopulation() {
+  let total = 0;
+
+  for (const countryPopulation of populationByCountries) {
+    const data = countryPopulation.textContent.trim();
+    const population = +parseInt(removeCommas(data));
+
+    total += population;
+  }
```
DanilWeda, line 19: In this case better to use reduce, because it`is typical case where we using this method, and you can practice it

DanilWeda (changes requested):
Good job! 
Check some comments.
Many thanks!

DanilWeda (approved):
Great!
Approved.
Many thanks)

PR #1183, opened 2024-06-04
Implement task solution

deezhizyu:


volodymyr-soltys97 (changes requested):
Hi, there is no demo link, also you need to add demo link to the PR description 
image

lerastarynets (approved):
Keep up the great work!

PR #1190, opened 2024-06-05
add solution

Nazarin565:
- [DEMO LINK](https://Nazarin565.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,23 @@
 'use strict';
 
-// write your code here
+const population = document.querySelectorAll('.population');
+
+const peoples = [];
+
+for (let i = 0; i < population.length; i++) {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,23 @@
 'use strict';
 
-// write your code here
+const population = document.querySelectorAll('.population');
+
+const peoples = [];
+
+for (let i = 0; i < population.length; i++) {
+  if (!isNaN(Number(population[i].textContent.split(',').join('')))) {
```
DarkMistyRoom, line 8: would be better to create a variable for this value. We don't need to calculate it twice
`const countryPopulation = Number(population[i].textContent.split(',').join(''));`

DarkMistyRoom (approved):
Good job! Only one suggestion

PR #1241, opened 2024-06-14
Develop

Karanelus:
- [DEMO LINK](https://karanelus.github.io/js_get_data_DOM/)
   
1. Get all text data from `span` with class `population`
2. Make sure that the given string can be converted to a number and convert it to number.
3. Calculate average and total value-based to parsed numbers.
4. Replace `Calculate it!` with calculated average numbers in `span` with classes `average-population` and `total-population`.
5. Calculated `average` and `total` numbers must be separated by a [thousands separator](https://www.codingem.com/comma-thousand-separator-in-javascript/)(same numeric style as populations presented).


BogdanMaliuta (approved):
Nice solution!

PR #1247, opened 2024-06-16
add task solution

AnnBusya:


lerastarynets (approved):
You're doing brilliantly, keep it up!

PR #1331, opened 2024-07-15
add task solution

Julia-na:
https://<Julia-na>.github.io/js_get_data_DOM/


lerastarynets (approved):
You're doing amazing, keep it up!

PR #1342, opened 2024-07-19
add solution

Valentyn-Radobenko:
In the end I realized that I could have created a single function and used it to generate a string for both total and average in the required format, but I've already done almost identical code for them and I only have two variables here, so it's not a big deal, but for the future I will think about it at the beginning


Valentyn-Radobenko:
- [DEMO LINK](https://Valentyn-Radobenko.github.io/js_get_data_DOM/)

PR #1355, opened 2024-07-23
Develop

coldshein:
- [DEMO LINK](https://coldshein.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,18 @@
 'use strict';
 
 // write your code here
+const collection = document.querySelectorAll('.population');
+const population = [...collection].map((item) => {
+  item.innerText.split(',').join('');
+});
+
+const totalPopulation = population.reduce((initial, item) => {
+  return +initial + +item;
+}, 0);
+const avgPopulation = totalPopulation / population.length;
+
+const avgPopulationElement = document.querySelector('.average-population');
+const totalPopulationElement = document.querySelector('.total-population');
+
+avgPopulationElement.innerText = avgPopulation.toLocaleString('en-US');
+totalPopulationElement.innerText = totalPopulation.toLocaleString('en-US');
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -7,7 +7,7 @@ const population = [...collection].map((item) => {
 });
 
 const totalPopulation = population.reduce((initial, item) => {
-  return +initial + +item;
+  return initial + Number(item);
 }, 0);
 const avgPopulation = totalPopulation / population.length;
 


src/scripts/main.js
```diff
@@ -1,3 +1,18 @@
 'use strict';
 
 // write your code here
+const collection = document.querySelectorAll('.population');
+
+const population = [...collection]
+  .map((item) => item.innerText.split(',').join(''))
+  .map((item) => Number(item))
+  .filter((item) => !isNaN(item)); // Переконуємося, що це число
```
BudnikOleksii, line 9: - `map` doesn't mutate an array, so you don't need to create a copy
- You can prepare one `map`
- Don't leave comments, especially in other language except for English, and this comment looks like ChatGPT leaves it

BudnikOleksii (changes requested):
Good work 👍 
Let's improve your solution slightly

src/scripts/main.js
coldshein: > * `map` doesn't mutate an array, so you don't need to create a copy
> * You can prepare one `map`
> * Don't leave comments, especially in other language except for English, and this comment looks like ChatGPT leaves it

that map is not for a reason of creating copy, just a way to iterate a NodeList, and back to the comment, I thought that leaving comments is a good idea for writing and learning but I changed my mind and forgot about that one 😅

BudnikOleksii: > > * `map` doesn't mutate an array, so you don't need to create a copy
> > * You can prepare one `map`
> > * Don't leave comments, especially in other language except for English, and this comment looks like ChatGPT leaves it
> 
> that map is not for a reason of creating copy, just a way to iterate a NodeList, and back to the comment, I thought that leaving comments is a good idea for writing and learning but I changed my mind and forgot about that one 😅

I mean you copy an array with a spread operator, it doesn't make sense as `map` doesn't mutate an array

src/scripts/main.js
```diff
@@ -1,3 +1,17 @@
 'use strict';
 
 // write your code here
+const collection = document.querySelectorAll('.population');
+
+const population = [...collection]
+  .map((item) => Number(item.innerText.split(',').join('')))
```
BudnikOleksii, line 7:
```suggestion
const population = collection
  .map((item) => Number(item.innerText.split(',').join('')))
```
I mean you don't need to create a copy with a spread operator

src/scripts/main.js
```diff
@@ -1,3 +1,17 @@
...
+  .filter((item) => item !== null);
```
BudnikOleksii, line 8: - You convert values to numbers, so why do you check if some of them are `null`?
- And, you don't need to strictly compare with null, you can just check if the value exists. https://developer.mozilla.org/en-US/docs/Glossary/Falsy

BudnikOleksii (changes requested):
Almost done, just a few more fixes are needed

src/scripts/main.js
coldshein: > I mean you don't need to create a copy with a spread operator
But NodeList haven't a method like a `map`, that's why I converting this collection into array with spread operator

BudnikOleksii: > > I mean you don't need to create a copy with a spread operator
> 
> But NodeList haven't a method like a `map`, that's why I've converted this collection into array with spread operator

Ah, got it, I forgot that

BudnikOleksii (approved):
👍

PR #1384, opened 2024-08-04
add task solution

VictorKomara:


src/scripts/main.js
```diff
@@ -1,3 +1,43 @@
...
+    i += 3;
+  }
+
+  const result = arrResult.join(',');
+
+  return result;
+}
+
+totalPopulation.innerHTML = getNumberSeparatedByCommas(total);
+averagePopulation.innerHTML = getNumberSeparatedByCommas(average);
```
etojeDenys, line 43: you can just use `.toLocaleString("en-US");`

etojeDenys (changes requested):

src/scripts/main.js
VictorKomara: Дякую за пораду

etojeDenys (approved):
lgtm

PR #1423, opened 2024-08-17
solution

yuraKrampus2021:
- [DEMO LINK](https://yuraKrampus2021.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,34 @@
 'use strict';
 
-// write your code here
+const dataSpans = document.querySelectorAll('.population');
+const totalElement = document.querySelector('.total');
+const averageElement = document.querySelector('.average');
+
+function removeCommas(str) {
+  return str.replace(/,/g, '');
+}
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,34 @@
...
-// write your code here
+const dataSpans = document.querySelectorAll('.population');
+const totalElement = document.querySelector('.total');
+const averageElement = document.querySelector('.average');
+
+function removeCommas(str) {
+  return str.replace(/,/g, '');
+}
+
+function formatNumberWithCommas(number) {
```
TarasHoliuk, line 11: That's nice you've implemented your own solution 👍 
So just FYI, there is a method that does the same (with the correct locale, but the default locale works as expected): [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString)

TarasHoliuk (approved):

PR #1424, opened 2024-08-19
add task solution

StanislavKapytsia:
Calculate average and total numbers and it must be separated by a thousands separator.

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,20 @@
 'use strict';
 
-// write your code here
+const ArrayOfSelectorPopulation = [...document.querySelectorAll('.population')];
+const total = document.querySelector('.total');
+const average = document.querySelector('.average');
+
+let newTotal = 0;
+let newAverage = 0;
+
+ArrayOfSelectorPopulation.forEach((selector) => {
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,20 @@
...
+const average = document.querySelector('.average');
+
+let newTotal = 0;
+let newAverage = 0;
+
+ArrayOfSelectorPopulation.forEach((selector) => {
+  const NUMBER = Number(selector.textContent.replaceAll(',', ''));
+
+  newTotal += NUMBER;
+});
```
etojeDenys, line 14: looks like you can simplify it using the reduce method

src/scripts/main.js
```diff
@@ -1,3 +1,20 @@
 'use strict';
 
-// write your code here
+const ArrayOfSelectorPopulation = [...document.querySelectorAll('.population')];
```
etojeDenys, line 3:
```suggestion
const arrayOfSelectorPopulation = [...document.querySelectorAll('.population')];
```
name vars and functions in camelCase

etojeDenys (changes requested):

VitaliyBondarenko1982 (approved):
GJ

PR #1426, opened 2024-08-21
Solution

DariaNastas:
[DEMO LINK](https://DariaNastas.github.io/js_get_data_DOM/)


etojeDenys (changes requested):
you should provide the demo link not the link to the repository

PR #1437, opened 2024-08-29
solution

Svidruk:


BudnikOleksii (changes requested):
Good work 👍 
Please, add a demo link so a reviewer can test your solution

src/scripts/main.js
```diff
@@ -1,3 +1,21 @@
...
+function toNumber(value) {
+  const number = parseInt(value.replaceAll(',', ''));
+
+  return isNaN(number) ? null : number;
+}
+
+const numbers = populationData.map(toNumber).filter((num) => num !== null);
+
+const total = numbers.reduce((acc, num) => acc + num, 0);
+const average = total / numbers.length;
```
VolodymyrKirichenko, line 18: Looks cool! 🔥 
But change the variable names, something like this. Use your own names
```suggestion
const populationElements = document.querySelectorAll('.population');
const totalPopulation = document.querySelector('.total-population');
const averagePopulation = document.querySelector('.average-population');

const populationData = Array.from(spans).map((span) => span.textContent);

function parsePopulation(value) {
  const number = parseInt(value.replaceAll(',', ''));

  return isNaN(number) ? null : number;
}

const populationNumbers = populationData.map(parsePopulation).filter((num) => num !== null);

const totalPopulation = numbers.reduce((acc, num) => acc + num, 0);
const averagePopulation = total / numbers.length;
```

VolodymyrKirichenko (approved):
Well done! 🥇

PR #1445, opened 2024-09-01
add solution

lerika7:
- [DEMO LINK](https://lerika7.github.io/js_get_data_DOM/)


etojeDenys (approved):
good job, but you can simplify that using the reduce method

PR #1446, opened 2024-09-04
Solution

kamilbykow:
- [DEMO LINK](https://kamilbykow.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,22 @@
 'use strict';
 
-// write your code here
+const docPopulationArray = document.querySelectorAll('.population');
+
+const docTotal = document.querySelector('.total-population');
+
+const docAverage = document.querySelector('.average-population');
+
+const populationArray = [];
+
+docPopulationArray.forEach((item) => populationArray.push(item.textContent));
+
+const populationNumbers = populationArray.map(
+  (item) => +item.replaceAll(',', ''),
+);
+
+const totalPopulation = populationNumbers.reduce((total, curr) => total + curr);
+
+const averagePopulation = totalPopulation / populationArray.length;
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,24 @@
...
+
+const docTotal = document.querySelector('.total-population');
+
+const docAverage = document.querySelector('.average-population');
+
+const populationNumbers = [];
+
+docPopulationArray.forEach((item) => {
+  populationNumbers.push(parseInt(item.textContent.replaceAll(',', '')));
+});
```
Zibi95, line 13: In JavaScript, the `map()` method is a built-in function commonly used to iterate over an array, applying a function to each element and returning a new array with the results. It does not modify the original array but instead creates a new one. However, it's not necessary to create a new array with elements parsed to numbers using `map()`. You can achieve this within the `reduce()` function as well.

src/scripts/main.js
```diff
@@ -1,3 +1,24 @@
...
+
+const totalPopulation = populationNumbers.reduce((total, curr) => total + curr);
+
+let averagePopulation;
+
+if (populationNumbers.length > 0) {
+  averagePopulation = totalPopulation / populationNumbers.length;
+}
```
Zibi95, line 21: so what if population.length === 0? I think it should be zero.
Instead of using let you can just use ternary 
```suggestion
const averagePopulation = populationNumbers.length > 0 ?  totalPopulation / populationNumbers.length : 0;
```

Zibi95 (changes requested):
Hey. Some things left to improve.

Zibi95 (approved):
LGTM!

PR #1451, opened 2024-09-06
solved

Artem-Brui:


src/scripts/main.js
```diff
@@ -1,3 +1,51 @@
...
+
+      return newResult;
+    }, '')
+    .split(',');
+
+  return separatedArray
+    .reverse()
+    .map((item) => item.split('').reverse().join(''))
+    .join(',');
+};
```
etojeDenys, line 45: too complex, you can just use `.toLocaleString()`

src/scripts/main.js
```diff
@@ -1,3 +1,51 @@
...
+  const customIndex = 1;
+
+  return array.reduce((total, item, index) => {
+    const currentFactor = Math.pow(1000, array.length - (customIndex + index));
+    const newTotal = total + Number(item) * currentFactor;
+
+    return newTotal;
+  }, 0);
+};
+
```
etojeDenys, line 20: and we also can simplify this one `parseInt(el.textContent.replace(/,/g, ''), 10)`

etojeDenys (changes requested):

src/scripts/main.js
Artem-Brui: thanks, really shorter :)

volodymyr-soltys97 (changes requested):
Hi, there is no demo link, you need to add demo link to the PR description 
image

etojeDenys (approved):

PR #1483, opened 2024-09-23
add solution

YuriiMykhailenko:
[DEMO LINK](https://YuriiMykhailenko.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,16 @@
 'use strict';
 
-// write your code here
+const countriesList = Array.from(document.querySelectorAll('span.population'));
+
+const totalPopulation = countriesList.reduce(
+  (acc, span) => acc + parseInt(span.textContent.split(',').join('')),
+  0,
+);
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,16 @@
 'use strict';
 
-// write your code here
+const countriesList = Array.from(document.querySelectorAll('span.population'));
+
+const totalPopulation = countriesList.reduce(
+  (acc, span) => acc + parseInt(span.textContent.split(',').join('')),
+  0,
+);
```
VolodymyrKirichenko, line 8: will be more readable
```suggestion
const totalPopulation = countriesList.reduce((acc, span) => (
  acc + parseInt(span.textContent.split(',').join(''))
), 0);
```

VolodymyrKirichenko (approved):
Well done 🔥

src/scripts/main.js
YuriiMykhailenko: Prettier thinks different😅

PR #1669, opened 2025-01-21
add task solution

Uusuff:


src/scripts/main.js
```diff
@@ -1,3 +1,16 @@
 'use strict';
+const COUNTRIES_BY_POPULATION = document.getElementsByClassName('population');
+const TOTAL_POPULATION = document.querySelector(".total-population")
+const AVERAGE_POPULATION = document.querySelector('.average-population');
```
etojeDenys, line 4: it's not constants, so they should be in camelCase

etojeDenys (changes requested):

PR #1760, opened 2025-03-05
Develop

fromjkeee16:
[DEMO LINK](https://fromjkeee16.github.io/js_get_data_DOM/)

### DIFF
--- a/src/scripts/main.js
+++ b/src/scripts/main.js
@@ -1,3 +1,25 @@
 'use strict';
 
-// write your code here
+const NUMBER_FORMAT = 'en-US';
+
+const items = [...document.querySelectorAll('.population')];
+
+const total = items.reduce(
+  (accumulator, element) => accumulator + parseNumber(element.textContent),
+  0,
+);
+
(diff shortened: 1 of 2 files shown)


src/scripts/main.js
```diff
@@ -1,3 +1,25 @@
...
+  getFormatted(average);
+
+document.querySelector('.total-population').textContent = getFormatted(total);
+
+function getFormatted(number, locale = NUMBER_FORMAT) {
+  return number.toLocaleString(locale);
+}
+
+function parseNumber(stringNumber, separator = ',', radix = 10) {
+  return parseInt(stringNumber.replaceAll(separator, ''), radix);
```
alexander-ignatow, line 24: Why do you pass radix? Do you expect the population to be passed in binary or hexadecimal form 😁 ?

alexander-ignatow (approved):
Good job!

src/scripts/main.js
fromjkeee16: its a general function, who knows the case...

PR #1937, opened 2025-07-09
finished task

ghost:
Finished countries population task


ghost (commented):
Fix typo: ent → document in querySelector

PR #1983, opened 2025-08-12
add task solution

TatyukOlena:


volodymyr-soltys97 (changes requested):
Hi, here is no demo link. Also, you need to provide the demo link to the PR description 
image

